PR #14, opened 2014-10-20
Delete 2003 files, no-one uses 2003 any more and one less thing to maint...

mikeparker:
...ain


squeek502:
I thought I remember Linux builds using the 2003 projects in some way.

mikeparker:
oh :( Any way we can get them to use 2005? Would be great to have one set of files to maintain.

squeek502:
Maybe? There's a lot about Linux that we probably need to re-figure out.

See: https://github.com/fortressforever/fortressforever/blob/beta/linux_ff/Makefile#L14

squeek502:
:+1: with my monkeying.

Linux is a lost cause for the moment. Can't do anything unless Valve gives us some binaries to work with.

PR #15, opened 2014-10-20
Remove unused string class, add folders to structure the solution a bit

mikeparker:
- Moved all the base hl2 stuff into a folder called /lib/
- Added folders for /projectiles/ /weapons/ and /grenades/ in shared, and /lua/ on server


squeek502:
Seems reasonable. Why `lib` instead of something like `SDK`?

squeek502:
Should probably either bring the 2008 project files in sync with these changes or delete them entirely. Does the 2008 sln serve a purpose?

mikeparker:
SDK might make more sense yeah.

I dunno about 2008, I can't remember who used it, was it Dexter? He's only using 2005 now so I don't know if something about 2008 broke or what. It's not exactly difficult to port a 2005 sln to 2008 at some point in the future anyway, so it just saves us time now removing it.

mikeparker:
Weird - my sln compiled with the string class deleted, i guess it needed a full rebuild? Better look out for that..

I've also added "fx" and "hud" folders for client

squeek502:
Well, judging by how old the 2008 vcproj files were (still had references to things like the vert grenade when I converted things from svn to git), I doubt anyone has used them in a very long time. Should just remove them.

squeek502:
:+1:

PR #23, opened 2014-10-22
Remove the speed check in IsInWorld for grenades/projectiles

### LINKED ISSUE #22: Grenades disappear after falling too far
From http://forums.fortress-forever.com/project.php?issueid=400

> Old bug, remember it existing way back when I just started playing ff. Usually happens when you throw a conc downwards from high altitudes. Fix plz?

Video demonstration: https://www.youtube.com/watch?v=QJixt671vA8

squeek502:
- Fixes #22: Grenades disappearing after falling too far

### DIFF
--- a/dlls/baseentity.h
+++ b/dlls/baseentity.h
@@ -946,7 +946,7 @@ public:
 	virtual bool	IsTemplate( void ) { return false; }
 	virtual bool	IsBaseObject( void ) const { return false; }
 	bool			IsBSPModel() const;
-	bool			IsInWorld( void ) const;
+	virtual bool	IsInWorld( void ) const;
 
 	// If this is a vehicle, returns the vehicle interface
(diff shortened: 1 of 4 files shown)


squeek502:
This seems to work fine. Can throw a conc straight down and it will fall until it explodes.

squeek502:
Now also works with pipes/IC rockets (it used to throw DataTable out-of-range value warnings)

mikeparker:
Looks ok. It can be seen as a bit of a smell to copy / paste code from base classes and override it, but I'm unsure what else we can do in this situation to prevent the base code from doing something we don't like other than editing it or avoiding calling it.

:+1:

squeek502:
Yeah, I didn't want to edit the base class directly as that could have unintended/unknown consequences (CBaseEntity is the root of pretty much everything).

PR #25, opened 2014-10-22
Remove unused 2008 sln/vcproj

squeek502:
Pretty sure no one uses these, and anyone that wants a 2008 sln can convert from the 2005 sln. Gives us one less thing to keep synchronized.


mikeparker:
:+1:

PR #26, opened 2014-10-22
Airshot/bounced rail death icons

squeek502:
Railgun icons
Airshot icons

From the commit messages:
- Most non-hitscan weapons will add the DMG_AIRSHOT flag to its DamageInfo's damagetype whenever an airshot happens (must be a direct hit)
- ConVar ffdev_airshot_height_threshold controls how high above the ground a player must be to be able to be counted as an airshot
- Airshots show in the console like so: `player killed player with weapon (airshot)`
- Airshots are logged to file like so: `"player<id><steamid><team>" killed "player<id><steamid><team>" with "weapon" (modifier "airshot")`
  - The `(modifier "")` string only gets added if there actually was an airshot, otherwise the entire parenthetical is absent from the line


mikeparker:
Is that SG image triggered by the SG rocket or bullets?

mikeparker:
Also is that a sniper airshot? Does that really count? - its the only one that is hitscan.

squeek502:
SG rockets only. I don't see any reason not to include sniper airshots; the time between shots and the damage of each shot makes them significant enough to broadcast in my opinion.

mikeparker:
Ok, sounds good :+1:

PR #28, opened 2014-10-22
Better info_ff_script bounds

squeek502:
Flags getting stuck in ceilings was caused by the flag bounds being 1 unit high while moving, then momentarily having 0 velocity while inside a ceiling, which caused its bounds to get reset and would get it stuck.

Walking right through thrown flags without catching them was caused by the 1 unit high flag size also being applied to the trigger/touch size. The touch size would only get set back to full once the flag velocity reached zero.

If I remember right, this decouples the trigger size from the physics size, allowing for a 1 unit high physics size while maintaining a constant trigger size. It also puts the control of the size in the hands of Lua, meaning that resupply bags/etc (which are also info_ff_scripts) can have different trigger bloat than flags.

From the commit message:
- Fixed flags getting stuck in ceilings
- Fixed flag touch bounds being 1 unit high while the flag is moving (it used to only get reset to the correct size once it had 0 velocity)
- Added support for separate collision and touch bounds for info_ff_scripts, but it's kind of a hacky way to do it (set size before physics calculations and then reset it after). The sizes are now controlled by Lua functions:
  - <entity_name>:gettouchsizes( mins, maxs ) passes min and max vectors by reference for the function to alter in place
  - <entity_name>:getphysicssizes( mins, maxs ) passes min and max vectors by reference for the function to alter in place
  - <entity_name>:getbloatsize() expects a float return value, uses the value to inflate the touch bounding box (default bloat is 12)

### DIFF
--- a/dlls/ff/ff_item_flag.cpp
+++ b/dlls/ff/ff_item_flag.cpp
@@ -39,7 +39,10 @@ extern "C"
 
 #include "tier0/memdbgon.h"
 
-#define ITEM_PICKUP_BOX_BLOAT		12
+ConVar ffdev_visualize_infoscript_sizes( "ffdev_visualize_infoscript_sizes", "0", FCVAR_FF_FFDEV );
+#define VISUALIZE_INFOSCRIPT_SIZES ffdev_visualize_infoscript_sizes.GetBool()
+
+#define ITEM_PICKUP_BOX_BLOAT 12 // default; only used if no lua function exists
 //ConVar ffdev_flag_throwup( "ffdev_flag_throwup", "1.6", FCVAR_FF_FFDEV );
 #define FLAG_THROWUP 1.6f
 //ConVar ffdev_flag_float_force( "ffdev_flag_float_force", "1.05", FCVAR_FF_FFDEV );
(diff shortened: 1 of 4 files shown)


mikeparker:
Sounds good. A while ago I think we saw that we couldn't reduce the height of the flag "grab" height without scaling the entire flag grab box down. Is it now possible to reduce the grab height of the flag without reducing the grab width?

squeek502:
Yeah, Lua has complete control. From the updated base_teamplay.lua:

``` lua
function baseflag:gettouchsize( mins, maxs )
    mins.x = mins.x / 2
    mins.y = mins.y / 2
    maxs.x = maxs.x / 2
    maxs.y = maxs.y / 2
    mins.z = 0
    maxs.z = maxs.z / 2
end

function baseflag:getphysicssize( mins, maxs )
    mins.x = mins.x / 2
    mins.y = mins.y / 2
    maxs.x = maxs.x / 2
    maxs.y = maxs.y / 2
    mins.z = 0
    maxs.z = 1
end

function baseflag:getbloatsize()
    return 0
end
```

mikeparker:
Ok awesome, :+1: for this

Issue #30, opened 2014-10-23
Nerf or remove the jump pad


mikeparker:
I guess this might be too big an issue to raise here? But I think a lot of us have come to the conclusion that the jump pad has run it's course and has more of a detrimental effect on both pickup / competitive games (CTF) and pub games (making AvD stacked too much in offense favour).

DexterHaslem:
a lot of us? I have to seriously ask because it's the first I've heard of it. but I dont really keep up on things and mostly just pub

squeek502:
We've discussed it a few times on the dev forums:
- [The jumppad](http://forums.fortress-forever.com/showthread.php?t=19837)
- [The Jump Pad: Options](http://forums.fortress-forever.com/showthread.php?t=21501)

AfterShock's pros/cons list from 2009 (before jump pads were made team-oriented/destructable):

> Jumppad pros:
> - It enables new players to easily access the fast movement and aircontrol that is core to FF
> - It allows scouts to improve their team and speed others up, not just themselves
> - A small amount of strategy (?)
> - Enables heavier offy classes than before (?)
> 
> cons:
> - Makes medics much faster than before
> - doesnt appear to encourage heavy offy classes as much as we'd like (?)
> - Is quite overpowered for AvD maps as the whole team can use it all the time
> - Makes some front line defense positions near impossible to defend (e.g. monkey T)
> - Places much more emphasis on back line defense as front line is easier to get past

My current thoughts:

I think if we really want to attempt to solve the "competitive FF is offense-biased" problem, the jump pad really needs to be looked at, as I think it's a major contributor for ~~two~~ three reasons:
- It allows for near complete consistency in runs that would otherwise be somewhat difficult/special (for example, rampsliding the FD lip on schtop/monkey)
- It allows medics to pull off the above mentioned runs with the same exact consistency as scouts, even where it'd otherwise be impossible (medics wouldn't be able to achieve scout-level conc speeds without a jump pad).
- It allows for traversing some yards very quickly without the need of a conc, meaning that either 1) a conc is saved where it'd otherwise be used or 2) offense runs are much faster where players would otherwise bhop across the yard

Something I've thought about which could solve the second problem (but not the first, and only partially the third) would be to make the jump pad boost based on the speed of the class (meaning scout would get the highest boost, and HW would get the lowest). However, this is not a great solution, as it devalues other reasons for the jump pad existing and may not solve much in the long run.

[...]

PR #33, opened 2014-10-25
Reduce jump-pad horizontal push

mikeparker:
https://github.com/fortressforever/fortressforever/issues/30
The idea here is to keep the good stuff about the jump pad:
- Give new players enough air time to easily access the "movement skill" part of FF, get used to air control, etc
- Make the scout less of a lone wolf and tie him to the team a little more
- Give the scout a purpose on AvD maps

Whilst reducing the impact:
- Reduce the impact of the jump pad on AvD especially on public servers
- Reduce the impact and usefulness of the jump pad on CTF in pickup / competitive situations

I have exposed the cvar in case anyone wants to experiment for themselves. I found 600 to be a good speed because:
- It's faster than a medic/scout bhop, so as a medic, it's worth taking
- It's slow enough that you shouldnt take a massive detour to use it

### DIFF
--- a/dlls/ff/ff_mancannon.cpp
+++ b/dlls/ff/ff_mancannon.cpp
@@ -45,8 +45,8 @@ extern const char *g_pszFFManCannonSounds[];
 
 extern const char *g_pszFFGenGibModels[];
 
-//ConVar ffdev_mancannon_push_forward( "ffdev_mancannon_push_forward", "1024", FCVAR_FF_FFDEV_REPLICATED );
-#define MANCANNON_PUSH_FORWARD 1024 //ffdev_mancannon_push_forward.GetInt()
+ConVar ffdev_mancannon_push_forward( "ffdev_mancannon_push_forward", "700", FCVAR_FF_FFDEV_REPLICATED );
+#define MANCANNON_PUSH_FORWARD ffdev_mancannon_push_forward.GetInt()
 //ConVar ffdev_mancannon_push_up( "ffdev_mancannon_push_up", "512", FCVAR_FF_FFDEV_REPLICATED );
 #define MANCANNON_PUSH_UP 512 //ffdev_mancannon_push_up.GetInt()
 //ConVar ffdev_mancannon_health( "ffdev_mancannon_health", "150", FCVAR_FF_FFDEV_REPLICATED );
--- a/dlls/ff/ff_mancannon.cpp
+++ b/dlls/ff/ff_mancannon.cpp
@@ -45,7 +45,7 @@ extern const char *g_pszFFManCannonSounds[];
 
 extern const char *g_pszFFGenGibModels[];
 
-ConVar ffdev_mancannon_push_forward( "ffdev_mancannon_push_forward", "700", FCVAR_FF_FFDEV_REPLICATED );
+ConVar ffdev_mancannon_push_forward( "ffdev_mancannon_push_forward", "600", FCVAR_FF_FFDEV_REPLICATED );
 #define MANCANNON_PUSH_FORWARD ffdev_mancannon_push_forward.GetInt()
 //ConVar ffdev_mancannon_push_up( "ffdev_mancannon_push_up", "512", FCVAR_FF_FFDEV_REPLICATED );
 #define MANCANNON_PUSH_UP 512 //ffdev_mancannon_push_up.GetInt()
--- a/game_shared/ff/ff_grenade_concussion.cpp
+++ b/game_shared/ff/ff_grenade_concussion.cpp
@@ -43,7 +43,7 @@
 // #0001629: Request: Dev variables for HH conc strength |-- Defrag
 
 //ConVar ffdev_mancannon_conc_speed( "ffdev_mancannon_conc_speed", "1700.0", FCVAR_FF_FFDEV_REPLICATED, "Max conc speed a player can attain after just using a jump pad." );
-#define MAX_JUMPPAD_TO_CONC_SPEED 1700.0f // ffdev_mancannon_conc_speed.GetFloat()
+#define MAX_JUMPPAD_TO_CONC_SPEED 1400.0f // ffdev_mancannon_conc_speed.GetFloat()
 //static ConVar ffdev_conc_lateral_power( "ffdev_conc_lateral_power", "2.74", FCVAR_FF_FFDEV, "Lateral movement boost value for hand-held concs", true, 0.0f, true, 2.74f );
 #define FFDEV_CONC_LATERAL_POWER 2.74f //ffdev_conc_lateral_power.GetFloat() //2.74f
 //static ConVar ffdev_conc_vertical_power( "ffdev_conc_vertical_power", "4.10", FCVAR_FF_FFDEV, "Vertical movement boost value for hand-held concs", true, 0.0f, true, 4.10f );


squeek502:
Just want to note that this doesn't address either of the following:

> - It allows for near complete consistency in runs that would otherwise be somewhat difficult/special (for example, rampsliding the FD lip on schtop/monkey)
> - It allows medics to pull off the above mentioned runs with the same exact consistency as scouts, even where it'd otherwise be impossible (medics wouldn't be able to achieve scout-level conc speeds without a jump pad).

which both stem from the conc+jump pad combo. We could address these by altering [MAX_JUMPPAD_TO_CONC_SPEED](https://github.com/fortressforever/fortressforever/blob/f663579932d1a685a2cb9195d3c25c3393c29b08/game_shared/ff/ff_grenade_concussion.cpp#L46) and potentially making it based on the class's speed (to allow conc + jump pad to have a higher limit for scouts than medics).

mikeparker:
To address the valid concern there, I've reduced the "jumppad-to-conc" speed across the board - so using a jump pad then a conc is worse than a good HH conc now. I think this is OK, as it was never designed to be combo'd, it was designed to be a (worse) accessible alternative, it should never be simply better than a conc, although at least it is more consistent. I hope this will mean people will put the skill back in HH concs over schtop.

mikeparker:
Pinging @squeek502 for thoughts

squeek502:
Wanted to test this locally first. Will comment later tonight.

squeek502:
This is a pretty brutal nerf. Effective, but brutal. It will probably obsolete the jump pad for competitive play, but I worry about it being obsoleted in public play as well, as it doesn't seem like there will be too many use cases for it anymore with the push being so weak (can barely rampslide from the boost) and it being destructible.

So, I'm not sure what I think. Can a jump pad that has been nerfed in this way justify its own existence? [This post by GreenMushy seems relevant](http://forums.fortress-forever.com/showpost.php?p=444393&postcount=26):

> I wrote a blurb about this in one of my posts a while back about making the positives about FF clearer to see. I think that is the whole truth behind the creation of the jumppad. It is an easier way for players to experience fast aerial movement without having to learn to conc. That way, it can plant the seed of teaching players why FF is cool by helping emphasize "HEY THIS GAME IS ABOUT MOVEMENT". I believe that the other (secondary) reasons for the jumppad were to evolve TFC, by implementing something cool and new, perhaps give small strategy, and try to introduce more team oriented mechanisms to the game. The notion that the jumppad is intended to make heavier classes more viable on O, i think, is an unintended side effect, and common misconception that makes people believe it is supposed to have as big an influence in organized play as it does on pubs.

As an aside, the jump pad has become FF's teleporter, considering how long its been around (competitve maps have been built with jump pads in mind [tidalwave yard would be the biggest example]).

mikeparker:
> It will probably effectively obsolete the jump pad for competitive play

I think that’s OK.

> I worry about it being obsoleted in public play as well, as it doesn't seem like there will be too many use cases for it anymore with the push being so weak (can barely rampslide from the boost) and it being destructible.

I could see it still being used in AvD, since it is still faster than bhop for every class. Slower classes will get a fairly solid boost, and there is a reasonable upwards push on it, so there are probably several places you can jump your teammates up to on a bunch of maps. With up to 8-10 players using it, I think it’ll still provide a considerable boost to the team – but the downside is, you’ll have a scout on your team so you are effectively sacrificing a player (kind of). I guess one argument is that this situation isn’t really a good one to be in, whereby you’re forcing a teammate to play a useless(?) class to help out the team..

Also, for players that can’t bhop, this is still 150% faster than scout run speed, so maybe the push only feels weak because we play at the very fastest level? It might feel reasonable to your average pub HWGuy?

If we are to keep the jump pad, I’d like to think it is one of many tools available to the player, which can be called on in specific circumstances. “You know what might work here? A jump pad!” rather than something that will definitely be useful in every situation.

I think that to a new player, the nerfed jump pad does satisfy the “Mushy” condition – showing new players what FF is all about, providing an easy way to learn air control and get used to higher speeds than they’ve previously had. 600 speed isn’t really that weak compared to most games, and even in FF it’s 150% the base speed of the fastest class – it only feels weak because we’re used to flying around at 1400 speed. That said, this goal of “showing new players what FF is about” - I place this goal pretty low on the priorities list, especially if it unbalances the game in general for the long term. So I guess the question is: If we can’t fit the jump pad into competitive play, and its only use is introducing air control / speed to new players, and we need to nerf it into the ground to prevent unbalancing, is it worth keeping at all?

The problem really is that we don’t have such a thing as an “offense support” class, like the engineer is for defense. I could imagine a class which is designed to boost the other offense by building structures on the perimeter of bases, buffing, disrupting the front line, etc, and then the jump pad might feel somewhat reasonable on that class. The scout is quite far from that role, though, which is another reason why it doesn’t really fit.

I guess one last thing to mention is that before we had jump pads, there was some mild resentment towards players that insisted on going scout – they were selfish glory-seekers. They didn’t help the team by killing anyone, they just purely went for the flag leaving the rest of the team to do “the hard work”. Now at least the team gets a benefit from someone going scout, and I’ve heard “someone should scout” more than once when the team is 4 medics and a bit slow. Maybe that’s a side issue, though.

Largos:
I could live with a nerfed speed, but from 1400 to 600 feels like too much.  Why would a scout ever take it?  800 seems like a nice medium.  

And with the original jpad speed you opened up certain trimp opportunities that were fun for heavier classes.

I generally like tools that boost movement, but that's my jam.  Don't remove the jpad, but a small buff in speed is fine with me.  Still think scout should get SOME benefit from his own tool, so 600 seems a tad slow.

squeek502:
The 2.46 horizontal jump pad boost was 1024, not 1400.

I think you're right that we can make the boost a bit higher than 600, though. I'm going to push some dlls to the Steam beta with some ffdev cvars exposed so that we can play around with the values a bit, and I've found some values that I think might feel a bit better (768 horizontal push, with 1024 max horiz jumppad+conc speed and 768 max vertical jumppad+conc speed).

AfterShockFF:
Sounds good :-)

AfterShockFF:
As with the other pull request - feel free to take over this work and make a new PR / close this one if you wish - i'll be away for a week and a bit.

AfterShockFF:
Small comment - 

> 768 horizontal push, with 1024 max horiz jumppad+conc speed and 768 max vertical jumppad+conc speed

capping horiz/vertical speeds separately could result in identical, reliable concs after using the jump pad, which means timing of the conc would not affect your angle and horizontal speed - which seems to be a reduction in depth and skill. Could be a bad thing.

squeek502:
> capping horiz/vertical speeds separately could result in identical, reliable concs after using the jump pad, which means timing of the conc would not affect your angle and horizontal speed - which seems to be a reduction in depth and skill. Could be a bad thing.

This is true as long as you're hitting the vertical speed cap every time, which is not the case (there is still a lot of vertical speed variability as long as you don't jump pad + conc as close together as possible). Also, with a low horizontal speed cap (like 1024), being able to get a consistent boost is not necessarily a bad thing as you'd (knowingly) be trading speed for consistency.

However, the reason I implemented separate horiz/vert caps is because it allows for a lower horizontal cap than an overall velocity cap would, because even though the "average" horizontal speed from a velocity cap like 1400 is ~1000 or so, you can still take advantage of the system by timing your conc so that it goes off when you're close to 0 vertical velocity, thereby getting close to the full 1400 speed as horizontal velocity.

AfterShockFF:
@squeek502 do you want to paste me your code changes or push what you did? I can change the 600 to 768 no problem, you implemented 2 separate conc-jpad caps though so I don't know whether to just reimplement that.

squeek502:
Here's my changes: https://github.com/fortressforever/fortressforever/tree/balance/jumppad-nerf

Let me know how you want to handle this.

AfterShockFF:
Open a PR with yours and I'll close this one.

PR #34, opened 2014-10-25
batch file to copy dlls into steam dir

mikeparker:
Will need editing on most peoples machines i guess

Is this a good way to do this? Anyone got better ideas?


squeek502:
Definitely need a more universal system to justify it being committed to the repo. I'd just leave this as a local file, and maybe put the code in the README as an example.

:-1:

AdamWillden:
I am also for putting this in the readme (my steam is on the G:/)

mikeparker:
I'll leave this out then, i've actually copied it outside my git directory so it doesn't disappear when i switch branch anyway

PR #36, opened 2014-10-25
Reduce skim - bhop_cap_hard reduced from 2.0 to 1.9

mikeparker:
This reduces the skim after a conc slightly. This means after conc skim 1 they are travelling at 684 speed rather than 720. This is actually 171% of movement speed rather than 190% due to the 0.9 degen multiplier but still a fair bit higher than TFC which nerfs you back to 100% speed after a conc.

### DIFF
--- a/cl_dll/ff/ff_hud_speedometer.cpp
+++ b/cl_dll/ff/ff_hud_speedometer.cpp
@@ -26,7 +26,7 @@
 
 // ELMO *** 
 #define BHOP_CAP_SOFT 1.4f // as defined in ff_gamemovement.cpp
-#define BHOP_CAP_HARD 2.0f // as defined in ff_gamemovement.cpp
+#define BHOP_CAP_HARD 1.9f // as defined in ff_gamemovement.cpp
 // *** ELMO
 
 // memdbgon must be the last include file in a .cpp file!!!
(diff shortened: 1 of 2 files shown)


squeek502:
Would be good to clean up the #defines/cvars here, as the "degen" setting does absolutely nothing except obscure the real cap.

Would be good to have a setting for the cap and a setting for the result of the cap (meaning if the cap was set to 1.9 and the result was set to 1.2, then if you jump while going >= 190% speed, you get set to 120% speed), just to keep our options open.

AdamWillden:
Hmm, sorry I copied the definitions I think that was one of my first changes to game code (or maybe I'm just a doofus), least I put a note! :)

What is that degen doing exactly?

squeek502:
The degen is an artifact from a previous implementation, before it was a _true_ hard cap.

mikeparker:
Done - i've removed the weird degen variable and adjusted the hard cap accordingly. I've added a hard cap result variable which is the same as the cap itself right now.

squeek502:
Nice. How do we want to handle things like this which might need value tweaking after its merged?

:+1: for testing this, at least.

mikeparker:
If we want to tweak a feature after it's merged we should definitely push changes to the same branch, then we decide to rebuild a beta / release without particular features we can just not include that branch. The question is whether to take the original branch out or not. To do it "properly", we should (take the old branch out) then put the new branch in, instead.

Issue #10, opened 2014-10-19
Fix ff_training loading from the main menu


squeek502:
I could not figure out how to start training while initiating the standard map loading screen. Instead, it simply runs the console command `map ff_training`, which is confusing to the user as the menu just disappears and you see nothing except the menu background image while loading.

Need to figure out how to programmatically start a local server on a certain map which triggers the standard map loading screen.

squeek502:
Fixed by 88b80666c461135545f434cee4a24cbfa907192b

PR #44, opened 2014-11-03
fixed ff_training not showing map loading info

alaswell:
See #10 just need to add 
engine->ClientCmd("progress_enable 1\n");


squeek502:
Wow, nice work. Thank you for looking into this. Did not expect it to be this simple. :+1:

Out of curiosity, how'd you figure it out?

alaswell:
I did too; I was trying to track down what the code for create server does and found the command in some random thread deep in the source coding forums.

PR #46, opened 2014-11-07
Flag/entity trails

mikeparker:
I think this is ready to go in now. It needs a corresponding Lua change in base_teamplay.lua

```
function baseflag:spawn()
//...
    flag:StartTrail(self.team)
end
```


hlstriker:
Maybe expose attaching a trail to any entity via lua. Also you shouldn't need to create the member variables using CHandles.

mikeparker:
ping @squeek502 for thoughts

One thing is that the red trails look a bit like grenade trails - any ideas as to how to reduce this? Maybe tweak the colours? Maybe its not an issue since it'll largely be coming from players?

squeek502:
Finally took a look at this.
- I'm unsure of the gameplay effects of having a trail on the flag specifically, but I like the idea of being able to add trails to any entity with Lua, so I'm all for this (the usage of it in baseflag would be a separate thing)
- Lua should have total control of the lifetime/size/color of the trail, with fallback/helper functions/constants for ease-of-use (see [CustomColor](https://github.com/fortressforever/fortressforever/blob/beta/dlls/ff/ff_lualib_misc.cpp#L40-L48), which exposes the Color class to Lua [and [see HudBox for an example of exposed functions that take Colors](https://github.com/fortressforever/fortressforever/blob/beta/dlls/ff/ff_lualib_globals.cpp#L1833-L1857)] and [the HudColors constants](https://github.com/fortressforever/fortressforever/blob/beta/dlls/ff/ff_lualib_constants.cpp#L192-L207))

mikeparker:
Ok - will add more customisability support. Ideally I'd like to be able to switch them off client side - it would be nice to have to give people the option. I can't think of a method of doing this without turning off all lua trails (which might kind of suck). Any thoughts?

squeek502:
Yeah, Lua and client-side toggleability don't really mix. Some possibilities:
- cvar toggle for trails in general (cl_lua_trails 0)
- separate cvar toggles for specific entity types (cl_lua_trails_info_ff_script 0)
- blacklist cvar for entity types (cl_lua_trails_blacklist_types "info_ff_script,prop_physics")
- blacklist cvar for entity names (cl_lua_trails_blacklist_names "some_name,*_flag")

mikeparker:
is info_ff_script only used for flags?

squeek502:
No, info_ff_scripts are simply generic point entities that have some special Lua callbacks and whatnot. All backpacks, health kits, powerups, whatever are also info_ff_scripts. It's the entity name (defined in hammer) that distinguishes them (and even then, that's not a required part of flags, that's just what base_teamplay uses).

There's not much in the C++ code that you can do to differentiate a flag from any other info_ff_script afaik.

squeek502:
An alternative solution:
- Allow Lua to define new client/server-side cvars

It's something I've thought would be cool/useful for a while, but never got around to doing. Sourcemod plugins can do this, so it should be possible. It's a more complicated solution, but it'd be way more robust.

EDIT: Actually, it'd still be difficult to implement client-side toggling, wouldn't it? You'd have to allow Lua to define which cvar applies to which trail and then use C++ to read the client-side value of the cvar and turn the trail on/off per-client.

AfterShockFF:
For now I'm thinking
- cvar toggle for trails in general (cl_lua_trails 0)

Might be best purely because its the least amount of work right now and we can always extend the functionality later.
(we might want to think up a name that ties more closely to what users might expect, but maybe thats difficult given the possibilities with lua)

AfterShockFF:
Ok so it looks like i've done this all server side and i'll need to send the message to clients instead, if i'm going to get per-client cvars working. Is UserMessages a bad way of doing this? Is there a better way?

AfterShockFF:
Have opened https://github.com/fortressforever/fortressforever/pull/144 for the new code

PR #48, opened 2014-11-08
Added sparks to the feet of rampsliders

mikeparker:
Kind of a bit of a hacky method to detect whether a player is rampsliding. If anyone has a better method, let me know. Currently its:
- If your ground entity is null AND
- If doing a trace straight downwards 0.1 units hits the floor AND
- If you are moving above 550 speed AND
- You are on an upwards slope

I wasn't sure whether to attach a sound also. It might be kind of cool to have a subtle sound when a player is rampsliding - currently you get complete silence, which is a bit weird?

### DIFF
--- a/game_shared/ff/ff_gamemovement.cpp
+++ b/game_shared/ff/ff_gamemovement.cpp
@@ -18,6 +18,7 @@
 #include "in_buttons.h"
 #include "movevars_shared.h"
 #include "ff_mapguide.h"
+#include "IEffects.h"
 
 #define	STOP_EPSILON		0.1
 #define	MAX_CLIP_PLANES		5
@@ -85,6 +86,7 @@ public:
 	virtual void WalkMove();
 	virtual void AirMove();
 	virtual void Friction();
+	bool IsRampSliding( CFFPlayer *pPlayer );
 
 	CFFGameMovement() {};
 };
@@ -901,6 +903,42 @@ bool CFFGameMovement::CanAccelerate()
 	return true;
 }
 
+bool CFFGameMovement::IsRampSliding( CFFPlayer *pPlayer )
+{
+	if (pPlayer->GetGroundEntity() == NULL)  // Rampsliding occurs when normal ground detection fails
+	{
+		// Take the lateral velocity
+		Vector vecVelocity = mv->m_vecVelocity * Vector(1.0f, 1.0f, 0.0f);
+		float flHorizontalSpeed = vecVelocity.Length();
+
+		if (flHorizontalSpeed > SV_TRIMPTRIGGERSPEED)
+		{
+			trace_t pm;
+
+			Vector vecStart = mv->m_vecAbsOrigin;
+			Vector vecStop = vecStart - Vector(0, 0, 0.1f);
+			
+			TracePlayerBBox(vecStart, vecStop, MASK_PLAYERSOLID, COLLISION_GROUP_PLAYER_MOVEMENT, pm); // but actually you are on the ground
+
+			// Found the floor
+			if(pm.fraction != 1.0f)
+			{
+				if (flHorizontalSpeed > 0)
+					vecVelocity /= flHorizontalSpeed;
+
+				float flDotProduct = DotProduct(vecVelocity, pm.plane.normal);
+				if (flDotProduct < -0.15f) // On an upwards ramp
+				{
+					return true;
+				}
+			}
+		}
+		
+	}
+
+	return false;
+}
+
 //-----------------------------------------------------------------------------
 // Purpose: Check player velocity & clamp if cloaked
 //-----------------------------------------------------------------------------
@@ -914,7 +952,12 @@ void CFFGameMovement::CheckVelocity( void )
 	CFFPlayer *pPlayer = ToFFPlayer( player );
 	if( !pPlayer )
 		return;
-	
+#ifdef CLIENT_DLL
+	if (IsRampSliding(pPlayer))
+	{
+		g_pEffects->Sparks(pPlayer->GetAbsOrigin());
+	}
+#endif
 	if( !pPlayer->IsCloaked() )
 		return;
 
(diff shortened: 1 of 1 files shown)


squeek502:
Rampsliding is actually an engine quirk that is only related to a vertical velocity threshold. See [these lines](https://github.com/fortressforever/fortressforever/blob/beta/game_shared/gamemovement.cpp#L3323-L3336), which are the totality of the rampsliding state (if going up at > 180 velocity, you're assumed to be in the air no matter what). This might not be new information for you, just wanted to clarify that.

Anyway, your conditions seem to make sense, as they better represent what people would recognize as rampsliding than the simple vertical velocity threshold.

Haven't checked it out ingame yet, but I like the idea. I also agree that a sound would be cool.

squeek502:
> 1:42 AM - FDA: what about when they're ramp sliding on cement? or dirt or wood?
> 1:42 AM - squeek.: sparks for everyone
> 1:43 AM - FDA: can you do a material specific particle system?
> 1:43 AM - squeek.: yeah

Might be a cool idea.

EDIT: It doesn't work all that well using the default impact effect stuff. Tested using the following code:

``` c++
#ifdef CLIENT_DLL
    #include "fx_impact.h"
#endif
```

``` c++
trace_t pm;

Vector vecStart = mv->m_vecAbsOrigin;
Vector vecStop = vecStart - Vector(0, 0, 0.1f);

TracePlayerBBox(vecStart, vecStop, MASK_PLAYERSOLID, COLLISION_GROUP_PLAYER_MOVEMENT, pm); // but actually you are on the ground

surfacedata_t *psurfaceData = physprops->GetSurfaceData( pm.surface.surfaceProps );
int iMaterial = psurfaceData->game.material;

PerformCustomEffects(pPlayer->GetFeetOrigin(), pm, Vector(0, 0, -1.0f), iMaterial, 1, 0);
```

Problems: some surfaces don't have material data (like metal grate textures). Concrete/etc has inconsistent effect starting locations; even if you spawn the effect at the feet, it (most of the time) will show at the center of the player. Unsure if it's worth it.

game_shared/ff/ff_gamemovement.cpp
```diff
@@ -914,7 +952,12 @@ void CFFGameMovement::CheckVelocity( void )
 	CFFPlayer *pPlayer = ToFFPlayer( player );
 	if( !pPlayer )
 		return;
-	
+#ifdef CLIENT_DLL
+	if (IsRampSliding(pPlayer))
+	{
+		g_pEffects->Sparks(pPlayer->GetAbsOrigin());
```
squeek502, line 958: This should use pPlayer->GetFeetOrigin()

squeek502:
Tried some tweaks to the sparks. I quite like something like this:

``` c++
Vector vecDir = -mv->m_vecVelocity / (mv->m_vecVelocity.LengthSqr() / 1000.0f);
g_pEffects->Sparks(pPlayer->GetFeetOrigin() + Vector(random->RandomFloat(-8, 8), random->RandomFloat(-8, 8), 0), 1, 1, &vecDir);
```

The sparks fly in the opposite direction that the player is moving and the origin is a bit randomized. Can also tweak the `1, 1` parameters to get bigger/longer sparks.

mikeparker:
Sounds good - i'll check it out. I wasn't aware of the 180 vertical speed thing - I expected you might know the rampslide code better than me!

mikeparker:
Out of curiosity, how are you testing it? I'm using thirdperson on ff_aardvark but I wonder if theres a way to see it better?

squeek502:
Some stuff I was doing
- `thirdperson` with `cam_idealdist 200`
- `host_timescale .25` occasionally
- changing the conditional to `if (true || IsRampSliding(pPlayer))`

mikeparker:
With the random origin, is there a chance that the sparks move into the ramp itself and are thus invisible?

squeek502:
It seems like there would be that chance but it doesn't seem to affect things as far as I could tell. I'm not sure that the sparks care about starting out inside of an entity.

mikeparker:
I'm just wondering if we could achieve what you want by altering the origin along the plane they are sliding on and whether that would be better? Might involve some maths, though.. :)

squeek502:
Something that might be useful would be to be able to get the trace back from `IsRampsliding`, so that we can use the trace data to do things like check the material and spawn dust instead of sparks if the material is dirt and react to the angle of the surface, etc.

Something like

``` c++
trace_t CFFGameMovement::FindGround( CFFPlayer *pPlayer )
{
    // trace and return the trace
}

bool CFFGameMovement::IsRampsliding( CFFPlayer *pPlayer, trace_t &tr)
{
    // check tr
}
```

``` c++
void CFFGameMovement::CheckVelocity( void )
{
    trace_t pm = FindGround(pPlayer);
    if (IsRampsliding( pPlayer, pm )
    {
        surfacedata_t *psurfaceData = physprops->GetSurfaceData( pm.surface.surfaceProps );
        int iMaterial = psurfaceData->game.material;

        if ( ( iMaterial == CHAR_TEX_DIRT ) || ( iMaterial == CHAR_TEX_SAND ) )
        {
            // dust
        }
        else
        {
            // sparks
        }
    }
}
```

DexterHaslem:
well if you're already considering tracing to figure out the material, might as well get dot of the plane and player vector and spew sparks from that angle at tr origin?

mikeparker:
I'll look into all this - all good ideas!

squeek502:
This currently only works for the client rampsliding, not for other people observing that client. Will either need to be put into a networked player var (m_bIsRampsliding) or the IsRampsliding check will need to be moved to somewhere that will get run for other players (gamemovement only runs for the local client afaik).

AfterShockFF:
Looks like sparks are triggered server side for SG sparks its definitely possible to send effect triggering over the wire. This may just work pulling out the client_dll ifdef. However, I suspect it'll be more optimal for each client to calculate it itself though, which would prevent loads of sparks messages getting sent.

squeek502:
Or would it be better for the server to calculate whether or not each player is rampsliding and then send that state to clients using a networked boolean?

AfterShockFF:
That would at least be cheap in terms of bandwidth i guess

AfterShockFF:
I'm away this weekend so I probably won't get to looking at this for another 2 weeks or so - feel free to give this a go yourself otherwise I'll likely get round to it when i get back.

AfterShock2:
Closing - squeek has taken this over #96

PR #53, opened 2014-11-10
linux build uses vc2005 project

### LINKED ISSUE #16: Replace / tweak vprojtomake so we can ditch 2003 sln/projs
Life would be easier if everyone (including linux) just used the 2005 files. This should be fairly easy, we just need to upgrade valves vprojtomake util so it reads 2005 instead of 2003 files.

The old one is here
https://github.com/fortressforever/fortressforever/tree/beta/utils/vprojtomake
The old instructions are here
https://developer.valvesoftware.com/wiki/Compiling_under_Linux
And the new 2010 (backwards compatible) version is here
https://developer.valvesoftware.com/wiki/Vprojtomake_2010

Apparently valve have (accidentally?) removed the hlds.so linux so we can only compile under windows right now. Squeek mentioned he might try to talk to a NeoTokyo dev to see what they did.

DexterHaslem:
- delete the 2003 project again RIP
- linux needs the memoverride-vc7 to work correctly. i renamed it to linux and included it in the project,
  only active w/ a linux preprocessor directive. the opposite is true for memoverride.cpp,
  This effectively lets us get away with updating a single project file, hooray.
- xerces xml crap removed from makefile and no longer needed on build machine w/ this change
- still not actually tested. might be gay

fixes #16


squeek502:
Nice.

PR #54, opened 2014-11-10
Allow the speedometer to show while spectating someone

### LINKED ISSUE #51: Make speedometer show while spectating
It's crazy that this isn't already the case.

squeek502:
- Closes #51

### DIFF
--- a/cl_dll/ff/ff_hud_speedometer.cpp
+++ b/cl_dll/ff/ff_hud_speedometer.cpp
@@ -56,7 +56,7 @@ public:
 		SetParent( g_pClientMode->GetViewport() );
 
 		// Hide when player is dead
-		SetHiddenBits( /*HIDEHUD_HEALTH |*/ HIDEHUD_PLAYERDEAD | HIDEHUD_NEEDSUIT );
+		SetHiddenBits( HIDEHUD_PLAYERDEAD );
 	}
 
 	virtual ~CHudSpeedometer( void )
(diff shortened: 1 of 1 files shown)


cl_dll/ff/ff_hud_speedometer.cpp
```diff
@@ -126,22 +126,26 @@ void CHudSpeedometer::OnThink()
...
-		Vector vecVelocity = local->GetAbsVelocity();
+		Vector vecVelocity = pPlayer->GetAbsVelocity();
 		m_iVelocity = (int) FastSqrt(
 			vecVelocity.x * vecVelocity.x 
 			+ vecVelocity.y * vecVelocity.y );
 
-		// average[i+1] = (average[i]*i + value[i+1])/(i+1)
-		m_flAvgVelocity = (m_flAvgVelocity*m_iNumUpdates + (float)m_iVelocity)/(m_iNumUpdates+1);
-		m_iNumUpdates++;
+		if (pPlayer == pLocalPlayer)
```
mikeparker: What does this equality mean?

cl_dll/ff/ff_hud_speedometer.cpp
```diff
@@ -126,22 +126,26 @@ void CHudSpeedometer::OnThink()
 	if (!hud_speedometer.GetBool() && !hud_speedometer_avg.GetBool())
 		return;
 
-	C_BasePlayer *local = C_BasePlayer::GetLocalPlayer();
+	C_FFPlayer *pPlayer = C_FFPlayer::GetLocalFFPlayerOrObserverTarget();
+	C_FFPlayer *pLocalPlayer = C_FFPlayer::GetLocalFFPlayer();
```
mikeparker: Whats the difference between these two variables?

cl_dll/ff/ff_hud_speedometer.cpp
squeek502: It checks that the player that we're getting the speed of (which can either be me or the person I'm spectating in first/third-person) is actually me. I considered  storing it in a named boolean, but I couldn't come up with a more descriptive variable name.

cl_dll/ff/ff_hud_speedometer.cpp
squeek502: Not sure what you're confused about here; the function names seem to make the difference obvious. The first will get your observer target if you're spectating in first/third-person, otherwise it'll get the local player. The second will always get the local player.

mikeparker: >  the function names seem to make the difference obvious.

Well yes, but it's often nice if variable names make the difference obvious so you don't have to go back and read what the variable was actually assigned from.
The note was just largely because the 2nd variable is only used for checking if the 1st was the same as it. Might be more readable if there was only one player variable and an IsSpectating boolean?

squeek502: Good idea. Did done it.

mikeparker:
Nice :) :+1:

PR #56, opened 2014-11-10
automagically copy dlls to FF steam directory

### LINKED ISSUE #55: automatically copy dlls to installation directory (steam only) on build
in spirit of #34 wanted to figure out a way to do this generically. it took some finagling with postbuild, but the good news is that it requires no configuration. strongly considering doing same thing for debugging

DexterHaslem:
used some powershell glue to dig it out of registry. please let me know if you have any issues with the powershell policy crap

closes #55 (and #34 really)

### DIFF
--- a/copy_bins_steam.ps1
+++ b/copy_bins_steam.ps1
@@ -1,0 +1,17 @@
+﻿param(
+    [string]$new_dll = ''
+)
+
+#1>new_dll = d:\fortressforever_myfork\dlls\Debug FF\2005\server.dll
+
+$steam_dir = (Get-ItemProperty HKCU:\Software\Valve\Steam).SteamPath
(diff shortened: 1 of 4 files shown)


copy_bins_steam.ps1
```diff
@@ -0,0 +1,17 @@
+﻿param(
+    [string]$new_dll = ''
+)
+
+#1>new_dll = d:\fortressforever_myfork\dlls\Debug FF\2005\server.dll
```
mikeparker, line 5: This looks suspiciously like a hardcoded user specific path - do we all need to do this manually or can we alter this to be automatic?

DexterHaslem, line 5: thats a comment i should have deleted. it uses the registry and vs output

PR #62, opened 2014-11-11
Lua: Various improvements

squeek502:
Some stuff requested by R00Kie.


squeek502:
Ready for review.

I wrote up a test script to ensure that everything functions as expected. The test definitions:

``` lua
local tests = {
    overheal = function(player)
        assert_equal( player:AddHealth(100, true), 100 )
    end,
    includescript = function(player)
        IncludeScript("subdir/test")
        assert_equal( subdir_file_loaded, true )
        subdir_file_loaded = nil
        IncludeScript("subdir\\test")
        assert_equal( subdir_file_loaded, true )
    end,
    isflashlighton = function(player)
        assert_equal( player:IsFlashlightOn(), true, "Flashlight needs to be on for this test" )
    end,
    isinair = function(player)
        assert_equal( player:IsInAir(), false, "Must be on the ground for this test" )
        assert_equal( player:IsInAir(32), false, "Overload should return the same thing" )

        local origin = player:GetOrigin()
        origin.z = origin.z + 16
        player:SetOrigin(origin)
        player:SetVelocity(Vector(0,0,100))

        delay(0.1, function()
            assert_equal( player:IsInAir(), true, "Should be off the ground" )
            assert_equal( player:IsInAir(32), false, "Should not be 32 units off the ground" )
            assert_equal( player:IsInAir(8), true, "Should be 8 units off the ground" )
        end)
    end,
    setdisguise = function(player)
        player:SetDisguise(Team.kRed, Player.kScout, true)
        delay(0.1, function()
            if player:GetClass() == Player.kSpy then
                assert_equal( player:IsDisguised(), true, "Spies can disguise" )
                player:ResetDisguise()
                assert_equal( player:IsDisguised(), false, "Spies can be undisguised" )
            else
                assert_equal( player:IsDisguised(), false, "Non-spies can not disguise" )
            end
        end)
    end,
    getfortpoints = function(player)
        local oldfortpoints = player:GetFortPoints()
        player:AddFortPoints(1, "Test")
        assert_equal( player:GetFortPoints(), oldfortpoints + 1 )
    end,
    getfrags = function(player)
        local oldfrags = player:GetFrags()
        player:AddFrags(1)
        assert_equal( player:GetFrags(), oldfrags + 1 )
    end,
    getdeaths = function(player)
        local olddeaths = player:GetDeaths()
        player:AddHealth(-1000)
        delay(0.1, function()
            assert_equal( player:GetDeaths(), olddeaths + 1 )
        end)
    end,
    getgamedescription = function(player)
        local olddesc = GetGameDescription()
        SetGameDescription("test")
        assert_equal( GetGameDescription(), "FF test" )
        SetGameDescription(olddesc)
    end,
    forcethrow = function(player)
        _G["entity"] = GetEntityByName("red_flag")
        _G["red_flag"]:touch(player)
        assert_equal(player:HasItem("red_flag"), true)
        ApplyToPlayer(player, {AT.kForceThrowItems})
        assert_equal(player:HasItem("red_flag"), false)
    end,
    forcedrop = function(player)
        _G["entity"] = GetEntityByName("red_flag")
        _G["red_flag"]:touch(player)
        assert_equal(player:HasItem("red_flag"), true)
        ApplyToPlayer(player, {AT.kForceDropItems})
        assert_equal(player:HasItem("red_flag"), false)
    end,
}
```

squeek502:
How are we going to handle things like this? It's all pretty straightforward/safe, but I can imagine its difficult to review without knowing how the Lua side of things works.

@AfterShockFF ?

dlls/ff/ff_lualib_player.cpp
```diff
@@ -25,6 +25,15 @@ extern "C"
 //---------------------------------------------------------------------------
 using namespace luabind;
 
+namespace FFLib
+{
+	// helper function to turn int into bool
```
AfterShockFF: This comment doesn't seem correct

dlls/ff/ff_scriptman.cpp
```diff
@@ -64,13 +64,6 @@ bool CFFScriptManager::LoadFile( lua_State *L, const char *filename)
 {
 	VPROF_BUDGET( "CFFScriptManager::LoadFile", VPROF_BUDGETGROUP_FF_LUA );
 
-	// don't allow scripters to sneak in scripts after the initial load
-	if(!_scriptman.m_isLoading)
-	{
-		Warning("[SCRIPT] Loading of scripts after initial map load is not allowed.\n");
-		return false;
-	}
```
AfterShockFF, line 72: Why was this block taken out, out of curiosity?

AfterShockFF:
Doesn't seem hard to review to me, I guess it depends what code review is expected to bring, though. A Lua layman won't be able to offer a full indepth review of your method and suggest that you do something completely different, but he can still spot typos, broken logic, styling / readability issues, etc.

Obviously it's hard to tell if your code simply doesn't work, from review, but that should be fairly obvious running it!

dlls/ff/ff_lualib_player.cpp
squeek502: It is correct. If you expose a function that takes a pointer as the first parameter, Luabind will treat it as if it was a member function of that class, so in this case it makes the exposed Lua function `bool CFFPlayer::IsFlashlightOn()` rather than `int CFFPlayer::IsFlashlightOn()`. When the int version was exposed, it would return a Lua number type (0 or 1) rather than a Lua boolean type (true or false).

dlls/ff/ff_scriptman.cpp
squeek502, line 72: It limits functionality for no good reason. See the commit message: https://github.com/fortressforever/fortressforever/commit/5083d5ccb7b78eb3f1ee19428998df8af0b3a1ba

squeek502, line 72: To be more specific, it stopped me from being able to test the IncludeScript changes after the initial map load, like so:

``` lua
    includescript = function(player)
        IncludeScript("subdir/test")
        assert_equal( subdir_file_loaded, true )
        subdir_file_loaded = nil
        IncludeScript("subdir\\test")
        assert_equal( subdir_file_loaded, true )
    end
```

dlls/ff/ff_lualib_player.cpp
AfterShockFF: Right, i see what you mean (assuming you mean CFFPlayer::FlashlightIsOn). My confusion was because a helper function to turn an int into a bool would be called something like IntToBool, not IsFlashlightOn! On closer inspection I can see the problem is that FlashlightIsIn returns an int rather than a bool. 

Minor - you could reword it to prevent dumbos like me getting confused:

 "// helper function since FlashlightIsOn annoyingly returns an int not a bool"

squeek502: Will do.

AfterShockFF:
:+1:

PR #67, opened 2014-11-16
Fix instances where '0' is used, but the null char is intended

squeek502:
- Fixes things like:

```
[SCRIPT] sv_mapluasuffix set to | finding maps\ff_2fort____.lua
[SCRIPT] maps\ff_2fort____.lua not found | reverting to maps\ff_2fort.lua
```


mikeparker:
lol, nice find :+1:

PR #68, opened 2014-11-16
Convert PostBuild.bat script to a pure batch script

squeek502:
[Good ol' stackoverflow](http://stackoverflow.com/questions/12069300/batch-file-discovered-program-path-variable-run-program-from-discovered-re)


mikeparker:
1 file seems better than 2  :+1:

squeek502:
It also avoids the PowerShell script permissions problem.

PR #75, opened 2014-12-01
Lua: Core improvements

squeek502:
The goal here is to make FF's Lua environment more robust while maintaining backwards compatibility.

The most important stuff so far (from a commit message):
- Backport a lot of the changes to the script manager made in the 2013 port
  - Make Lua's 'print' function redirect to the console
  - Add lua_dostring server command that will attempt to run the given string in the global Lua environment
  - Expose all safe default Lua library functions (newly exposed packages: package, debug, os)
  - Make Lua's 'require' look in ModDir/maps/includes, ModDir/maps/ and ModDir/ when resolving modules
- Remove unused CRC32 script validation system
- Standardize Lua console messages/warnings 

Some stuff that I'd like to do:
- [x] Deprecate and/or make `IncludeScript` use `require`
- [x] Add `tostring` support for userdata so that, at the very least, userdata prints something like `<entity_classname>:<memory_address>`


dlls/ff/ff_scriptman.cpp
```diff
@@ -33,51 +33,197 @@ extern "C"
...
+
+// redirect Lua's print function to the console
+static int print(lua_State *L)
+{
+	int n=lua_gettop(L);
+	int i;
+	for (i=1; i<=n; i++)
+	{
+	if (i>1) Msg("\t");
+	if (lua_isstring(L,i))
```
AfterShockFF: I'm guessing this code is ported from somewhere but it could benefit from a little cleanup:
- Indentation is off 
- "n" and "i" variables could be named better
- One-line "if" statements could be broken onto two lines and space out the expressions

Again, not a big deal though

AfterShockFF:
Fairly hard to digest but assuming it all works why not :+1:

dlls/ff/ff_scriptman.cpp
squeek502: Yeah, it was copy-pasted, and I plan on improving it. Right now it'll print all exposed C++ classes as simply `userdata:<pointer address>` which is not very useful. I'm going to investigate the best method for getting tostring to work properly for userdata.

squeek502:
Rebased and force pushed after merging #62 to avoid merge conflicts in ff_scriptman.cpp

PR #76, opened 2014-12-01
Fix ff_restartround resetting twice when called with 0 prematch time

squeek502:
- Also makes 0 prematch the default when using ff_restartround without a parameter (ff_setprematch is better for cases where you just want to add prematch)


AfterShockFF:
:+1:

PR #79, opened 2014-12-02
Lua: Add global function GetEntitiesByName(entname) that returns a Lua table containing the matching entities

squeek502:
A more straightforward alternative to the more complex Collection-based `Collection():GetByName({"table_of_names"})` method


AfterShockFF:
:+1:

PR #81, opened 2014-12-02
Fix conc speed being limited for too long after using a jump pad

### LINKED ISSUE #80: Jump pad + conc speed limiting persists for way too long
See: https://github.com/fortressforever/fortressforever/blob/beta/game_shared/ff/ff_grenade_concussion.cpp#L370

Currently, it'll limit your conc speed if you conc anytime within _5.2 seconds_ after using a jump pad. Meaning, you can use the jump pad, _then_ prime a conc, and when that conc goes off your speed will still be limited.

Some possible solutions (using both would probably be best):
- Reset m_flMancannonTime when the player jumps or IsOnGround
- Lower the time to something more reasonable

squeek502:
- Fixes #80
- Will stop limiting once the player is on the ground or in water, or after 3 seconds has passed since the jump pad was used
- Also made the jump pad set players as in the air, so that m_flMancannonTime doesn't have to be checked in CFFGameMovement::CheckJumpButton (being in the air will force an early return)

**Notable side effect:** This allows a semi-exploit in that you can use a jump pad, land, and jump + conc to get a hard-cap-limited, full speed conc. I'm not sure how to get around that, as the old behavior was definitely buggy (it doesn't make sense to limit conc speed once you've landed from the jump pad boost). At the very least, it's way more interesting/dynamic than the old jump pad + conc combo mechanic.

### DIFF
--- a/game_shared/ff/ff_gamemovement.cpp
+++ b/game_shared/ff/ff_gamemovement.cpp
@@ -79,6 +79,7 @@ public:
 	virtual bool CanAccelerate();
 	virtual void FullNoClipMove(float factor, float maxacceleration);
 	virtual void CheckVelocity( void );
+	virtual void CategorizePosition( void );
 
 	// CFFGameMovement
 	virtual void FullBuildMove( void );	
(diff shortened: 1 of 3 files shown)


AfterShockFF:
I think this is interesting, and maybe not a bad thing. It requires pretty exact timing. I'd be interested to try it.

game_shared/ff/ff_gamemovement.cpp
```diff
@@ -471,6 +457,21 @@ bool CFFGameMovement::CheckJumpButton(void)
 	return true;
 }
 
+void CFFGameMovement::CategorizePosition()
+{
+	BaseClass::CategorizePosition();
+
+#ifdef GAME_DLL
+	CFFPlayer *pFFPlayer = ToFFPlayer(player);
+	if (pFFPlayer->m_flMancannonTime > 0.0f && player->GetGroundEntity() != NULL || player->GetWaterLevel() > WL_Feet)
```
AfterShockFF, line 466: Remind me what mancannontime is, the time since you last used it?

AfterShockFF:
Code looks fine :+1:

game_shared/ff/ff_gamemovement.cpp
squeek502, line 466: It's the time that you last used a jump pad: https://github.com/fortressforever/fortressforever/blob/beta/dlls/ff/ff_mancannon.cpp#L217

PR #85, opened 2014-12-05
Scout movement weapon

squeek502:
We were pretty happy with this in the old beta; just needed to get it finalized. Cherry-picked all the relevant commits from the dev/svn branch (hopefully I didn't miss any).

Things left to do:
- [ ] Name it (its currently code named jumpdown because I'm a dumbo)
- [ ] Give it a weapon model (currently uses the railgun model)
- [x] Give it a unique sound (currently uses the railgun sound)
- [ ] Give it a better effect (currently uses a simple conc-like ring)

For anyone unfamiliar:
- When used, gives the scout a double-jump-like boost (moves them in the direction they are looking, not an additive boost)
- Must be fully charged to be used
- Charges passively while held, loses all charge when switching weapons

Some quotes of mine from the old beta forums:

> The goal is to communicate that the scout is primarily for movement. The current (2.4) weapon loadout does the exact opposite.
> 
> So, we basically have two options:
> - Remove all weapons, leaving the scout with "hands"
> - Add a movement weapon that requires you to have it out basically at all times except for certain cases (nailgunning an SG for instance)
> 
> We have done the latter, and I think it's a far better solution.

---

> The other thing is that it's not really meant to be the most useful thing ever. As long as it is used instead of damage weapons, the goal has been met.
> 
> The scout is for movement, that is all we want to communicate. If this weapon does that, then fantastic. 

---

> Anyone playing scout trying to get kills is not going to be successful, and will conclude (incorrectly) that the game is unbalanced because of the scout's perceived weakness. This movement weapon will just make sure that no longer happens; if people no longer play scout, that's their prerogative. At least we'll no longer see D scouts.
> 
> Also, scout has always been a "forced" role (in that his weapons are awful; the best scouts don't use weapons). We're just making sure people understand that role, because as of now, it's not easy to grasp with the clues FF gives (it's hard to tell that his weapons are intended to be awful, and that it's better if you avoid using them altogether). 

---

> If this gets any one person to think "oh, I shouldn't be killing people with this class", it will have succeeded. If it also doesn't unbalance the scout in CTF while at the same it time improves the scout's usefulness in AvD, it will have succeeded quadrupaley.

---

> Here are the goals I personally had in mind for a scout movement weapon:
> - Simple, accessible movement
> - Unique new movement tool (opens up new possibilities with movement, not just adding another way to achieve similar movement)
> - Not overpowering (or mostly doesn't affect the balance of the scout)
> - Can only be used for movement
> - Requires that it be used in place of damage weapons
> - Communicates that the scout shouldn't use damage weapons


AfterShockFF:
The only other thing we werent quite happy with, design mechanics wise, (if I remember correctly, was a while ago!) was switching weapons back and forth and whether you lose all your charge and stuff - I can't remember exactly what we decided.

game_shared/ff/ff_weapon_base.h
```diff
@@ -137,6 +137,10 @@ typedef enum
 	// Engineer specific
 	FF_WEAPON_RAILGUN, 
 
+	// Scout specific
+	FF_WEAPON_JUMPDOWN,
+	FF_WEAPON_JUMPUP,
```
AfterShockFF: i suspect we don't need both?

game_shared/ff/ff_weapon_jumpdown.cpp
```diff
@@ -0,0 +1,569 @@
...
+/// @file ff_weapon_jumpdown.cpp
+/// @author Gavin "Mirvin_Monkey" Bramhill
+/// @date December 21, 2004
+/// @brief The FF railgun code & class declaration
+///
+/// REVISIONS
+/// ---------
+/// Jan 19 2004 Mirv: First implementation
+//
+//	11/11/2006 Mulchman: Reverting back to bouncy rail
```
AfterShockFF: We should probably try to get rid of these useless comments on new files, at least we should try to not make the problem worse in the absence of a big cleanup.

game_shared/ff/ff_weapon_jumpdown.cpp
```diff
@@ -0,0 +1,569 @@
...
+		m_flStartTime = gpGlobals->curtime - JUMPDOWN_CHARGEUPTIME;
+		m_flLastUpdate = gpGlobals->curtime - JUMPDOWN_CHARGEUPTIME;
+		m_flTotalChargeTime = JUMPDOWN_CHARGEUPTIME;
+		m_flClampedChargeTime = JUMPDOWN_CHARGEUPTIME;
+
+		m_flNextPrimaryAttack = gpGlobals->curtime;
+	}
+
+	//Reset the last weapon to this jumpgun
+	ToFFPlayer(GetOwnerEntity())->SetLastFFWeapon(this);
```
AfterShockFF: I am somewhat surprised this must be done manually.

game_shared/ff/ff_weapon_jumpdown.cpp
```diff
@@ -0,0 +1,569 @@
...
+	CFFWeaponBase* pLastWeapon = ToFFPlayer(GetOwnerEntity())->GetLastFFWeapon();
+
+	//If the weapon is valid
+	if( pLastWeapon != NULL )
+	{
+		//If the last weapon was a jumppad, dont reset the jumpgun values
+		if( pLastWeapon->GetWeaponID() == FF_WEAPON_DEPLOYMANCANNON )
+		{
+			//The last weapon was a jumpad
+			DevMsg("Last weapon was a jumpad.  Do nothing\n");
```
AfterShockFF: We probably dont need these devmsgs any more

game_shared/ff/ff_weapon_jumpdown.cpp
```diff
@@ -0,0 +1,569 @@
...
+	{
+		DevMsg("Last weapon was NULL.  Deploy charged jumpgun\n");
+
+		m_flStartTime = gpGlobals->curtime - JUMPDOWN_CHARGEUPTIME;
+		m_flLastUpdate = gpGlobals->curtime - JUMPDOWN_CHARGEUPTIME;
+		m_flTotalChargeTime = JUMPDOWN_CHARGEUPTIME;
+		m_flClampedChargeTime = JUMPDOWN_CHARGEUPTIME;
+
+		m_flNextPrimaryAttack = gpGlobals->curtime;
+	}
```
AfterShockFF: if you pulled all the code above here out into a function "you could return early and reduce the nested logic (which makes it quite hard to work out whats going on). In fact - if you don't need the below line (SetLastFFWeapon), you could just return early from inside the logic here.

game_shared/ff/ff_weapon_jumpdown.cpp
```diff
@@ -0,0 +1,569 @@
...
+
+	//Reset the last weapon to this jumpgun
+	ToFFPlayer(GetOwnerEntity())->SetLastFFWeapon(this);
+
+//#else
+
+	//m_flTotalChargeTimeBuffered = m_flClampedChargeTimeBuffered = 0.0f;
+	//m_flChargeTimeBufferedNextUpdate = 0.0f;
```
AfterShockFF: can get rid of this stuff

game_shared/ff/ff_weapon_jumpdown.cpp
```diff
@@ -0,0 +1,569 @@
...
+	Vector vecSrc = pPlayer->Weapon_ShootPosition();
+	//Vector vecSrc = pPlayer->GetAbsOrigin() + vecForward * 8.0f + vecRight * 5.0f + Vector(0, 1, (pPlayer->GetFlags() & FL_DUCKING) ? 5.0f : 22.0f);
+
+	float flPercent = m_flClampedChargeTime / JUMPDOWN_CHARGEUPTIME;
+
+	// Push them
+	if (!JUMPDOWN_VERTICALSETVELOCITY && !JUMPDOWN_HORIZONTALSETVELOCITY)
+		pPlayer->ApplyAbsVelocityImpulse(Vector(horizPush.x, horizPush.y, JUMPDOWN_VERTICALPUSH) * flPercent);
+	else if (JUMPDOWN_VERTICALSETVELOCITY && JUMPDOWN_HORIZONTALSETVELOCITY)
+	    pPlayer->SetAbsVelocity(Vector(horizPush.x, horizPush.y, JUMPDOWN_VERTICALPUSH) * flPercent);
```
AfterShockFF: when we are happy with the jumpgun behaviour we can remove a lot of this testing-different-velocity-maths stuff (I suspect the time might be now).

game_shared/ff/ff_weapon_jumpdown.cpp
```diff
@@ -0,0 +1,569 @@
...
+
+	SendWeaponAnim( GetPrimaryAttackActivity() );
+
+	// MUST call sound before removing a round from the clip of a CMachineGun
+	WeaponSound(SINGLE);
+
+	// Player "shoot" animation
+	pPlayer->SetAnimation( PLAYER_ATTACK1 );
+	
+	int nShots = min(GetFFWpnData().m_iCycleDecrement, pPlayer->GetAmmoCount(m_iPrimaryAmmoType));
```
AfterShockFF: Does it even have ammo?

game_shared/ff/ff_weapon_jumpdown.cpp
```diff
@@ -0,0 +1,569 @@
...
+		JUMPDOWN_EFFECT_R,						//r
+		JUMPDOWN_EFFECT_G,						//g
+		JUMPDOWN_EFFECT_B,						//b
+		JUMPDOWN_EFFECT_ALPHA,					//a
+		JUMPDOWN_EFFECT_SPEED,					//speed
+		FBEAM_FADEOUT
+		);
+
+#ifdef GAME_DLL
+#endif
```
AfterShockFF: not very cool ifdef block

game_shared/ff/ff_weapon_jumpdown.cpp
```diff
@@ -0,0 +1,569 @@
...
+{
+
+	CFFPlayer *pPlayer = GetPlayerOwner();
+
+	if (!pPlayer)
+		return;
+
+#ifdef GAME_DLL
+	// Not currently charging, but wanting to start it up
+	if (m_flStartTime == -1.0f && pPlayer->GetAmmoCount(GetPrimaryAmmoType()) > 0)
```
AfterShockFF: More ammo stuff

game_shared/ff/ff_weapon_jumpdown.cpp
```diff
@@ -0,0 +1,569 @@
...
+	if (pPlayer->GetAmmoCount(GetPrimaryAmmoType()) <= 0 && m_flNextPrimaryAttack <= gpGlobals->curtime)
+	{
+		HandleFireOnEmpty();
+
+		// HEV suit - indicate out of ammo condition
+		pPlayer->SetSuitUpdate("!HEV_AMO0", FALSE, 0); 
+	}
+
+#ifdef GAME_DLL
+#else // ^^ GAME_DLL ^^
```
AfterShockFF: This can just be #ifdef CLIENT_DLL

AfterShockFF:
Looks ok, could do with a slight bit of cleanup. Probably want to rename the gun too.. :) :+1: merge when ready

AfterShockFF:
Is there a corresponding script change, presumably it just needs the scout with the gun added to his inventory, and the ff_weapon_jumpgun.txt?

squeek502:
Yeah, I assume there are a few script changes necessary. Weapon script file, scout player script file, sound script addition. Those changes should be buried somewhere in the old SVN branch. I'll try to dig them up and submit a corresponding PR on the scripts repo later tonight.

AfterShockFF:
Ok, thanks. If you have any thoughts on the weapon sound let me know otherwise I'll try to dig something out.

squeek502:
@BullyFF made something (see #115) but it's no longer online. I'll check if I downloaded it when I get a chance.

BullyFF:
@squeek502 @AfterShockFF it's here https://soundcloud.com/bullrushband/ffjumpgun1

squeek502:
Thanks @BullyFF 

@AfterShockFF, added what we had in the old beta to fortressforever/fortressforever-scripts#54. Will definitely need some fleshing out, though (given its own sound definition, adding a localized string for its name to the language files, etc).

FDA-:
If the jump gun is going to replace one of the other weapons I'd like to opt for the scout keeping his nailgun, I think it's weapon for team work allowing the scout to help with very weak sentries occasionally.

squeek502:
Definitely agree. In fortressforever/fortressforever-scripts#54 it's in slot3, not replacing any weapons.

AfterShockFF:
Going to merge this - it's releasable in this state. It doesnt have a unique weapons model but thats unlikely to come soon and we can do it later. I've added the sounds. Weapons naming and adding a better effect can always be added later.

arenaconspiracy:
i don't suppose you ever considered the jump gun being used on any other class while making it did you? i feel like it's an interesting concept to explore, and with a tad bit of tweaks could fit in the arsenal of for example sniper, with little changes to the weapon and medium (or heavy, you know.) changes to the class, it could find a new home in use.

PR #88, opened 2014-12-08
Generate <classname>.cfg/userconfig.cfg files at runtime if they don't exist

### LINKED ISSUE #87: Stop distributing classname.cfg/userconfig.cfg directly
Instead, these should be generated at runtime using a _default.cfg that we distribute (like how config.cfg is created from config_default.cfg). The files are too prone to being overwritten when we distribute them directly.

This should be done before we release on Steam.

squeek502:
- Closes #87 in conjunction with fortressforever/fortressforever-scripts#3

### DIFF
--- a/cl_dll/ff/ff_cdll_client_int.cpp
+++ b/cl_dll/ff/ff_cdll_client_int.cpp
@@ -1,0 +1,73 @@
+#include "cbase.h"
+#include "ff_cdll_client_int.h"
+
+#include "filesystem.h"
+#include "ff_utils.h"
+
+#define CLASSCFG_PATH				"cfg/%s.cfg"
+#define CLASSCFG_DEFAULT_PATH		"cfg/classcfg_default.cfg"
+#define USERCONFIG_PATH				"cfg/userconfig.cfg"
+#define USERCONFIG_DEFAULT_PATH		"cfg/userconfig_default.cfg"
+
+int CFFClient::Init( CreateInterfaceFn appSystemFactory, CreateInterfaceFn physicsFactory, CGlobalVarsBase *pGlobals )
+{
+	int ret = BaseClass::Init( appSystemFactory, physicsFactory, pGlobals );
+
+	PopulateMissingConfigs();
+	
+	return ret;
+}
+
+void CFFClient::PopulateMissingConfigs()
+{
+	// populate class configs
+	if (filesystem->FileExists(CLASSCFG_DEFAULT_PATH, "MOD"))
+	{
+		char szConfigPath[MAX_PATH] = {0};
+		CUtlBuffer buf(0,0,false);
+		bool bIsBufferPopulated = false;
+
+		for (int iClass=CLASS_SCOUT; iClass<=CLASS_CIVILIAN; iClass++ )
+		{
+			Q_snprintf(szConfigPath, sizeof(szConfigPath), CLASSCFG_PATH, Class_IntToString(iClass));
+
+			if (!filesystem->FileExists(szConfigPath))
+			{
+				if (!bIsBufferPopulated)
+				{
+					FileHandle_t f = filesystem->Open(CLASSCFG_DEFAULT_PATH, "rb");
+					
+					if ( f == FILESYSTEM_INVALID_HANDLE )
+						continue;
+
+					filesystem->ReadToBuffer( f, buf );
+					filesystem->Close(f);
+
+					bIsBufferPopulated = true;
+				}
+
+				filesystem->WriteFile(szConfigPath, "MOD", buf);
+			}
+		}
+	}
+
+	// populate userconfig (helper cfg for class cfgs)
+	if (filesystem->FileExists(USERCONFIG_DEFAULT_PATH, "MOD") && !filesystem->FileExists(USERCONFIG_PATH, "MOD"))
+	{
+		CUtlBuffer buf(0,0,false);
+		FileHandle_t f = filesystem->Open(USERCONFIG_DEFAULT_PATH, "rb");
+
+		if ( f != FILESYSTEM_INVALID_HANDLE )
+		{
+			filesystem->ReadToBuffer( f, buf );
+			filesystem->Close(f);
+
+			filesystem->WriteFile(USERCONFIG_PATH, "MOD", buf);
+		}
+	}
+}
+
+CFFClient gFFClient;
+IBaseClientDLL *clientdll = &gFFClient;
+
+EXPOSE_SINGLE_INTERFACE_GLOBALVAR( CFFClient, IBaseClientDLL, CLIENT_DLL_INTERFACE_VERSION, gFFClient );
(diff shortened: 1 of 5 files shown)


cl_dll/ff/ff_cdll_client_int.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+	PopulateMissingConfigs();
+	
+	return ret;
+}
+
+void CFFClient::PopulateMissingConfigs()
+{
+	// populate class configs
+	if (filesystem->FileExists(CLASSCFG_DEFAULT_PATH, "MOD"))
```
AfterShockFF: whats MOD here?

cl_dll/ff/ff_cdll_client_int.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+		for (int iClass=CLASS_SCOUT; iClass<=CLASS_CIVILIAN; iClass++ )
+		{
+			Q_snprintf(szConfigPath, sizeof(szConfigPath), CLASSCFG_PATH, Class_IntToString(iClass));
+
+			if (!filesystem->FileExists(szConfigPath))
+			{
+				if (!bIsBufferPopulated)
+				{
+					FileHandle_t f = filesystem->Open(CLASSCFG_DEFAULT_PATH, "rb");
```
AfterShockFF: whats "rb"?

cl_dll/ff/ff_cdll_client_int.cpp
```diff
@@ -0,0 +1,73 @@
...
+					if ( f == FILESYSTEM_INVALID_HANDLE )
+						continue;
+
+					filesystem->ReadToBuffer( f, buf );
+					filesystem->Close(f);
+
+					bIsBufferPopulated = true;
+				}
+
+				filesystem->WriteFile(szConfigPath, "MOD", buf);
```
AfterShockFF: This whole loop could do with a bit more readability help. Hungarian prefixes hurt when it gets to szBlah and we have a buf(0,0,false) (i have no idea what those zeros are, or what the false means). we have a variable called "f", and a couple of magic strings "MOD" and "rb".

cl_dll/ff/ff_cdll_client_int.cpp
```diff
@@ -0,0 +1,73 @@
...
+	{
+		char szConfigPath[MAX_PATH] = {0};
+		CUtlBuffer buf(0,0,false);
+		bool bIsBufferPopulated = false;
+
+		for (int iClass=CLASS_SCOUT; iClass<=CLASS_CIVILIAN; iClass++ )
+		{
+			Q_snprintf(szConfigPath, sizeof(szConfigPath), CLASSCFG_PATH, Class_IntToString(iClass));
+
+			if (!filesystem->FileExists(szConfigPath))
```
AfterShockFF: I think you can invert this "if" statement and just return early if the file doesnt exist, reduces nesting

cl_dll/ff/ff_cdll_client_int.cpp
```diff
@@ -0,0 +1,73 @@
...
+		}
+	}
+
+	// populate userconfig (helper cfg for class cfgs)
+	if (filesystem->FileExists(USERCONFIG_DEFAULT_PATH, "MOD") && !filesystem->FileExists(USERCONFIG_PATH, "MOD"))
+	{
+		CUtlBuffer buf(0,0,false);
+		FileHandle_t f = filesystem->Open(USERCONFIG_DEFAULT_PATH, "rb");
+
+		if ( f != FILESYSTEM_INVALID_HANDLE )
```
AfterShockFF: Again you could return early here and invert the if statement, reduces nesting

AfterShockFF:
This is generally a good idea. The code could do with a slight cleanup but otherwise all good :+1:

cl_dll/ff/ff_cdll_client_int.cpp
squeek502: "MOD" is a weird constant for telling the filesystem which search paths to check (options seem to be "MOD" and "GAME"). I can't really find much documentation about it, but I did find this in dlls/navfile.cpp:

```
    bool navIsInBsp = false;
    FileHandle_t file = filesystem->Open( filename, "rb", "MOD" );  // this ignores .nav files embedded in the .bsp ...
    if ( !file )
    {
        navIsInBsp = true;
        file = filesystem->Open( filename, "rb", "GAME" );  // ... and this looks for one if it's the only one around.
    }
```

We pretty much always use "MOD" in FF files.

cl_dll/ff/ff_cdll_client_int.cpp
squeek502: Seems like it takes the same mode flags as [fopen](http://www.cplusplus.com/reference/cstdio/fopen/) (including the 't' flag). So, 'read in binary mode', basically.

cl_dll/ff/ff_cdll_client_int.cpp
squeek502: Did some more research about search paths. They are defined in FileSystem_Steam.dll and set based on [gameinfo.txt](https://github.com/fortressforever/fortressforever-scripts/blob/master/gameinfo.txt#L24-L47). Although we can't access the implementation in that dll, there is an analog (which seems to be accurate) in [public/filesystem_init.cpp](https://github.com/fortressforever/fortressforever/blob/beta/public/filesystem_init.cpp#L498-L590) that's used to compile demoinfo.exe.

Fortress Forever's current search paths:

```
NULL pathID search paths:
c:\program files (x86)\steam\steamapps\common\fortress forever\platform\
c:\program files (x86)\steam\steamapps\common\fortress forever\fortressforever\
c:\program files (x86)\steam\steamapps\common\fortress forever\hl2\
c:\program files (x86)\steam\steamapps\common\fortress forever\platform\config\

GAME search paths:
c:\program files (x86)\steam\steamapps\common\fortress forever\fortressforever\
c:\program files (x86)\steam\steamapps\common\fortress forever\hl2\

MOD search paths:
c:\program files (x86)\steam\steamapps\common\fortress forever\fortressforever\

GAMEBIN search paths:
c:\program files (x86)\steam\steamapps\common\fortress forever\fortressforever\bin\
c:\program files (x86)\steam\steamapps\common\fortress forever\hl2\bin\

PLATFORM search paths:
c:\program files (x86)\steam\steamapps\common\fortress forever\platform\

DEFAULT_WRITE_PATH search paths:
c:\program files (x86)\steam\steamapps\common\fortress forever\fortressforever\
```

The 2013 engine uses search paths to allow customizations to not overwrite files, and instead they'd get loaded from `<gamedir>/custom`. We can sort of do the same thing, but 2013 has a [new filesystem_init.cpp implementation](https://github.com/ValveSoftware/source-sdk-2013/blob/55ed12f8d1eb6887d348be03aee5573d44177ffb/mp/src/public/filesystem_init.cpp#L531-L722) that allows it to be less intrusive (we'd have to set fortressforever/custom as the first mod/game path, but then various temp/cfg/etc files get written to it instead of where they should be; [2013 defines /custom/ as a special case using a new wildcard thingy](https://github.com/ValveSoftware/source-sdk-2013/blob/master/mp/game/mod_hl2mp/gameinfo.txt#L42-L49)).

This probably isn't the right place to post this, but oh well.

PR #89, opened 2014-12-09
Fixed CFFCrosshairOptions::UpdateSliders assert and did some cleanup

squeek502:


cl_dll/ff/vgui/ff_crosshairoptions.cpp
```diff
@@ -520,7 +523,7 @@ void CFFCrosshairOptions::UpdateCrosshairs()
 void CFFCrosshairOptions::UpdateSliders()
 {
 	int iCurrentWeapon = m_pWeapon->GetActiveItem();
-	Assert(iCurrentWeapon >= 0 && iCurrentWeapon <= FF_WEAPON_TOMMYGUN);
+	Assert(iCurrentWeapon >= 0 && iCurrentWeapon <= FF_HIT_CROSSHAIR);
```
AfterShockFF: this iCurrentWeapon is a bit confusing now - is this actually currentCrosshair ?

AfterShockFF:
Minor suggestion otherwise all good :+1:

cl_dll/ff/vgui/ff_crosshairoptions.cpp
squeek502: That m_pWeapon and iCurrentWeapon has always been incorrectly named, as 0 is actually the Global crosshair. I'll do another pass to clean it up some more.

PR #90, opened 2014-12-09
Default hud_fastswitch to 1 and add hud_weaponselect cvar

squeek502:
- See #4 and http://forums.fortress-forever.com/showthread.php?t=24632

> Maybe we need a middle ground that shows the fastswitch HUD component when changing weapons, but doesn't require you to press fire to actually switch. 

That's what this is.


AfterShockFF:
Sounds like a great compromise. :+1:

PR #96, opened 2014-12-14
Rampslide sparks (take 2)

squeek502:
Updated duplicate of #48, with some spark improvements and networking

Things that could/should be done:
- [x] Emit something other than sparks (dust) when sliding on grass/dirt/etc. This would probably entail moving all the tracing client-side and removing the networked var
- [x] Add client cvars to toggle rampslide sparks or control their size/etc

### DIFF
--- a/game_shared/ff/ff_gamemovement.cpp
+++ b/game_shared/ff/ff_gamemovement.cpp
@@ -18,7 +18,6 @@
 #include "in_buttons.h"
 #include "movevars_shared.h"
 #include "ff_mapguide.h"
-#include "IEffects.h"
 
 #define	STOP_EPSILON		0.1
 #define	MAX_CLIP_PLANES		5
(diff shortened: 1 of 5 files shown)


game_shared/ff/ff_gamemovement.cpp
```diff
@@ -907,6 +908,42 @@ bool CFFGameMovement::CanAccelerate()
 	return true;
 }
 
+bool CFFGameMovement::IsRampSliding( CFFPlayer *pPlayer )
+{
+	if (pPlayer->GetGroundEntity() == NULL)  // Rampsliding occurs when normal ground detection fails
+	{
+		// Take the lateral velocity
+		Vector vecVelocity = mv->m_vecVelocity * Vector(1.0f, 1.0f, 0.0f);
+		float flHorizontalSpeed = vecVelocity.Length();
```
AfterShockFF, line 917: Minor - its possible to reduce the nesting here quite a lot by inverting the ifs and returning false immediately, e.g.

```
if (pPlayer->GetGroundEntity() != NULL) // Rampsliding occurs when normal ground detection fails
    return false;

// Take the lateral velocity
Vector vecVelocity = mv->m_vecVelocity * Vector(1.0f, 1.0f, 0.0f);
float flHorizontalSpeed = vecVelocity.Length();
if (flHorizontalSpeed <= SV_TRIMPTRIGGERSPEED)
    return false;

// etc
```

AfterShockFF:
Other than a minor comment this is all good :+1:

BullyFF:
https://soundcloud.com/tudberry/sets/ffsoundstuff

squeek502:
Nice, thanks. Can you make them downloadable (in the Permissions tab of each track)?

BullyFF:
Oops! Done :)

AfterShock2:
@squeek502 is this ready to go in?

squeek502:
Yeah, adding it now would probably be fine. We can add sounds afterwards.

PR #97, opened 2014-12-14
Remove the map guide menu as it has never been utilized

### LINKED ISSUE #94: Modify the map guide options on team select.
- Basically no maps in FF have an image for the map guides. Therefore, there should probably either be a default one if one isn't found, or it should be blank rather than being the black and purple missing texture checkerbox.
- Most maps don't even have map guides. It'd probably be better if the non-existent mapguides are greyed out and unselectable.
  - And if there are no map guides at all, I'd recommend having the "View Flythrough" option on the team select screen greyed out and unselectable.

squeek502:
- Closes #94
- Also removed the FullNoClipMove implementation from ff_gamemovement.cpp because it was totally obsolete


AfterShockFF:
Fine with me :+1:

PR #122, opened 2015-03-22
Reduce jump-pad horizontal push (take 2)

squeek502:
See #33


AfterShockFF:
Do we want the vars exposed for release or do u want to hardcode them?

squeek502:
The [FCVAR_FF_FFDEV_REPLICATED define includes the FCVAR_CHEAT flag](https://github.com/fortressforever/fortressforever/blob/3ec0faa920834ec82441d2dd0538d9757199aff9/public/tier1/convar.h#L85), so it's okay to leave it in if we want.

AfterShockFF:
Ok looks good :+1:

PR #123, opened 2015-03-22
Optimise lua element strings to convert to integer IDs 

### LINKED ISSUE #18: Convert Lua HUD identifiers to a byte when sending over the network
Something that has been discussed elsewhere.

See: https://github.com/fortressforever/fortressforever/blob/beta/game_shared/ff/ff_utils.cpp#L831

No reason to send the entire string; the client has no use for it other than to identify which HUD element to update. A byte is sufficient because there is already a [hard limit defined of 128 HUD elements](https://github.com/fortressforever/fortressforever/blob/beta/cl_dll/ff/ff_hud_lua.h#L32).

AfterShockFF:
Closes #18


AfterShockFF:
letsago :+1:

PR #126, opened 2015-03-23
Lua: Majorly improve Lua's ability to interact with sentryguns

squeek502:
Also did some refactoring of CFFSentryGun so that the exposed functions are more useful. 
Closes #70

Player functions added to get the player's buildables (or nil if they're not built):

```
  player:GetSentryGun()
  player:GetDispenser()
  player:GetDetpack()
  player:GetJumpPad()
```

Sentry Gun functions added:

```
  sg:SetLevel(int level) // sets the level (does not play upgrade sounds)
  sg:Upgrade() // upgrades to the next level
  sg:Repair(int cells) // adds health based on the number of cells
  sg:AddAmmo(int shells, int rockets) // adds ammo
  sg:RocketPosition() // returns the Vector position of where rockets are fired from
  sg:MuzzlePosition() // returns the Vector position of where bullets are fired from
  sg:GetRockets() // number of rockets the SG has
  sg:GetShells() // number of shells the SG has
  sg:GetHealth() // amount of health the SG has
  sg:SetRockets(int rockets)
  sg:SetShells(int shells)
  sg:SetHealth(int health)
  sg:GetMaxRockets()
  sg:GetMaxShells()
  sg:GetMaxHealth()
  sg:SetFocusPoint(Vector point) // sets where the SG wants to look
  sg:GetEnemy() // gets the target of the SG (CBaseEntity)
  sg:SetEnemy(CBaseEntity enemy)
  sg:GetVecAiming() // gets the direction vector of the SG's current aim direction
  sg:GetVecGoal() // gets the direction vector of the SG's goal aim direction
  sg:Shoot() // makes the SG shoot one shell
  sg:ShootRocket() // makes the SG shoot one rocket
```

Test code that I was using:

``` lua
local sg = player:GetSentryGun()
assert_equal(player:GetSentryGun() ~= nil, true, "Player should have an SG (build one first)")
local max_shells = sg:GetMaxShells()
local max_health = sg:GetMaxHealth()
local max_rockets = sg:GetMaxRockets()
assert_equal(max_shells > 0, true, "Max shells should be greater than zero")

sg:AddAmmo(max_shells, max_rockets)
sg:Repair(max_health)
assert_equal(sg:GetShells(), max_shells, "SG should have max cells")
assert_equal(sg:GetHealth(), max_health, "SG should have max health")
assert_equal(sg:GetRockets(), max_rockets, "SG should have max rockets")

sg:SetLevel(1)
[...]

PR #127, opened 2015-03-23
Consolidate SG locked on status icon updates in CFFSentryGun::SetEnemy

squeek502:
- Fixes the status icon not being updated if Lua calls SetEnemy (see #126)


AfterShockFF:
Nice :) :+1:

Issue #91, opened 2014-12-09
Convert Lua Collections to use Lua tables instead of std::vector


squeek502:
The std::vector implementation is unnecessary and seems to have some issues (seen it cause a few hard-to-debug crashes related to iterators while destructing at Shutdown).

See: https://github.com/fortressforever/fortressforever/blob/beta/dlls/ff/ff_lualib_util.cpp

PR #129, opened 2015-03-23
Lua: Add global GetPlayers function

squeek502:
- Returns a table (array) of all players on the server

Should be used as a superior alternative to:

``` lua
local players = Collection()
players:GetByFilter( {CF.kPlayers} )
```

(Collections always loop through the entire entity list)

Slightly related to #91


AfterShockFF:
Looks good :+1:

PR #131, opened 2015-03-23
Lua: Remove luabind's 'class' implementation because it seems to be broken

squeek502:
It's not used anywhere. See: http://www.rasterbar.com/products/luabind/docs.html#defining-classes-in-lua


AfterShockFF:
ok :+1:

PR #132, opened 2015-03-24
Lua: Add IsEntity function

squeek502:
Tested using the following code:

``` lua
        assert_equal(IsEntity(player), true, "Player should be an entity")
        assert_equal(IsEntity(nil), false, "nil should not be an entity")
        assert_equal(IsEntity(function() end), false, "Function should not be an entity")
        assert_equal(IsEntity({}), false, "Empty table should not be an entity")
        assert_equal(IsEntity(true), false, "Boolean should not be an entity")
        assert_equal(IsEntity(10), false, "Number should not be an entity")
```


AfterShockFF:
:+1:

PR #134, opened 2015-03-24
Add various Lua improvements that aid in moving Collection to Lua

squeek502:
Some misc functions useful for fortressforever/fortressforever-scripts#27

PR #136, opened 2015-03-24
Remove the C++ Collection implementation

### LINKED ISSUE #91: Convert Lua Collections to use Lua tables instead of std::vector

squeek502:
- Closes #91

Heavily overlaps with (because it depends on) #134. Only 7a6b42ec0aa1332e9288fea2a16423212bbede92 is truly relevant to this PR.


AfterShockFF:
:+1:

Issue #121, opened 2015-03-22
Lua timers don't get their time synced properly


squeek502:
From fortressforever/fortressforever-scripts#21

The time gets set whenever the client receives the message from the server, which means that it will end up showing a different time than the server's. One fix might be to send gpGlobals->curtime along with the message and then have the client offset the start time based on the difference of the received curtime and their own curtime.

squeek502:
So, the issue that sparked this will be fixed by #137 (the timer appeared to be 1 second faster than what it should be). However, this still might be worth doing? I'm not sure.

Doing what I suggested in the OP, it turns out that in a listen server, the client gpGlobals->curtime is actually behind the server's when they receive the message (even with net_fakelag set to a high value). Would need to test the difference on a dedicated server with real ping to see if this is at all worth doing, but I'd imagine the difference is pretty negligible and not worth sending the extra float.

PR #137, opened 2015-03-24
Lua: Fix HUD timers' seconds getting truncated in the wrong direction when counting down

squeek502:
- Was causing 0.9999 seconds left to show as 0
- Timers counting up still have the same behavior as before (0.9999 seconds will show as 0)
- Potentially obsoletes #121


AfterShockFF:
:+1:

PR #145, opened 2015-03-27
Add lua-spawned trails on entities, coloured by team

AfterShockFF:
Needs a corresponding change in base_teamplay.lua:

https://github.com/fortressforever/fortressforever-scripts/pull/32


AfterShockFF:
I've added an overload so lua authors can easily create trails with sensible defaults. However if they want to know what the defaults are, purely for small tweaks, whats the best way of exposing this information?

squeek502:
:+1:

PR #172, opened 2015-03-30
Use the player's Steam Friends name as their default name

### LINKED ISSUE #167: It's time to update the default name.
We're on Steam. Lets do it how other games do it. Make the default name the player's steam name. But able to change it to anything else if they like.

https://partner.steamgames.com/documentation/api --> Persona name

squeek502:
- Closes #167

Will need the `name "WhenFFwillOut"` line removed from config_default.cfg


AfterShockFF:
Looks good :+1:

PR #173, opened 2015-03-30
Fix schedule/timer/menuman memory leak

squeek502:
24f8f8c6f2873dd0827411648aa6ee74478ef501 removed the deletes from CFFScheduleManager and claimed that CUtlMap's remove functions do the deletion, but I can find no evidence for that being the case

The only other instance of a dynamically allocated pointer being stored as the value of a CUtlMap is CPhysSaveRestoreBlockHandler::m_QueuedRestores, which does its own deletions. This commit mirrors CPhysSaveRestoreBlockHandler's logic exactly.

### DIFF
--- a/dlls/ff/ff_menuman.cpp
+++ b/dlls/ff/ff_menuman.cpp
@@ -333,7 +333,10 @@ void CFFLuaMenuManager::RemoveLuaMenu(const char* szMenuName)
 	// remove the schedule from the list
 	unsigned short it = m_menus.Find(id);
 	if(m_menus.IsValidIndex(it))
+	{
+		delete m_menus[it];
 		m_menus.RemoveAt(it);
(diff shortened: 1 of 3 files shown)


squeek502:
Note also that this could also be solved by removing the dynamic allocation itself and just store the objects in the map directly (would use the copy constructor when adding them to the map).

I ain't know which would be better.

AfterShockFF:
I'm surprised we need to do this, but if we do, this looks ok :+1:

squeek502:
I'm going to revert this because I've been getting some mysterious crashes while changing map that seem to be related to Lua shutting down and freeing memory. Reverting this locally seemed to fix those crashes as well.

I guess I shouldn't try to fix what I'm not totally sure is broken.

PR #175, opened 2015-03-30
Reduce grellow pipe damage radius to 125 (from 130)

### LINKED ISSUE #158: Potentially lower pipe radius
With the smaller flag bounds, pipe radius doesn't need to be so big anymore. It's currently set to 130. For reference, RPG radius is 100.

squeek502:
- TFC has a radius of ~120
- Closes #158 

Figured I'd at least make the pull request just in case we want to do it.

### DIFF
--- a/game_shared/ff/ff_weapon_pipelauncher.cpp
+++ b/game_shared/ff/ff_weapon_pipelauncher.cpp
@@ -27,7 +27,7 @@
 #endif
 
 //ConVar pipe_damage_radius( "ffdev_pipe_damage_radius", "130", FCVAR_FF_FFDEV_REPLICATED, "Pipe explosion radius" );
-#define PIPE_DAMAGERADIUS 130.0f //pipe_damage_radius.GetInt()
+#define PIPE_DAMAGERADIUS 125.0f //pipe_damage_radius.GetInt()
 
 //----------------------------------------------------------------------------
 // Purpose: Send special hint on firing first pipe


AfterShockFF:
I actually think TFCs radius was too small which led to not quite enough demomen (and variation in defense classes). We might want 125 or something in between. But I'm on board with reducing it slightly.

squeek502:
I agree, dropping it to 120 might be too much. A very slight decrease or no change at all would still be okay.

Jitterskull:
Eh, is it really necessary?? I didn't even find them to be this powerful. They still do damage and easier to hit than blue pipes, though. Basically, it can kill a scout in one blow while blue not while blue is good already the way it is. Not sure if reducing it's overall radius would help.

If we are going to change the demoman to lower his massive damage with dps and his huge AoE radius (which is good for house cleaning class anyway), personally I'd suggest to nerf his ridiculously powerful MIRV. I'm not talking about the 4 explosive pieces he drop, but the direct hit is imho waaay too powerful. The buff it once got wasn't really necessary, the explosion is so powerful it can almost literally kill a medic in one blow if it explodes near him or at least left him with 5hp.

squeek502:
I was curious about the buff you thought the MIRV got and looked through the code a bit. The MIRV damage has never been increased (it's been 180 since 1.0 and is still 180), but I did find that we intended to reduce both the MIRV and Mirvlet damage in 2.4 to match frag gren damage (145) but only actually reduced Mirvlet damage and left MIRV untouched by accident.

Will submit a PR to correct that in a bit.

Jitterskull:
Oh... after I checked the changelogs, I discovered that this is correct. When I made that post I had memories of much weaker MIRV, so I was probably thinking of some other thing (in most case, about the unofficial 1.12 FF version that at one point of time was ironically more popular than official FF release).

PR #177, opened 2015-03-30
Remove encryption key so .ctx player/weapon scripts don't get loaded

squeek502:
- Makes it so that .txt files are prioritized over .ctx, so that if the pre-Greenlight encrypted scripts still exist on a server install, they get ignored


AfterShockFF:
:+1:

PR #180, opened 2015-03-30
Fix spy disguise weapon for classes that got weapons removed in 2.5.0

### LINKED ISSUE #178: Spy weapon visual bug
While holding a gun in a slot that the disguised class doesn't have, it will show the spy's original weapon instead of something the class should be holding. It only happens on classes that recently had some weapons removed.

squeek502:
- Fixes tranq for medic/hwguy/pyro and nailgun for sniper
- Fixes #178


AfterShockFF:
nice, refactoring too! :) :+1:

Issue #186, opened 2015-04-05
Lua GetTeams() function


R00Ki3:
Could we add a GetTeams function that would return the active teams in an array?

squeek502:
Even better would be to allow Lua to do this directly if it's not already possible and then add the GetTeams function to includes/util/utils.lua. Functions needed for that as far as I can tell:
- [x] team:GetPlayerLimit()

What do you mean by active teams? Would those include unassigned/spectator?

R00Ki3:
Yes you could include unassigned and spec and basically any team that doesn't have SetPlayerLimit set to -1. Yeah, team:GetPlayerLimit() Would work the same and I could make a function to use it however i want.

PR #187, opened 2015-04-05
Lua: Add get equivalents of exposed team set functions

squeek502:
Adds functions:
 team:GetAllies() // returns a table of the team's allies (table of CFFTeam objects)
 team:GetPlayerLimit() // returns the team's player limit
 team:GetClassLimit(int class_id) // returns the team's class limit for the given class

Also allows GetTeam to return the unassigned/spectator teams.

Contributes towards #186 

Tested using the following code:

``` lua
assert_equal(GetTeam(Team.kUnassigned) ~= nil, true, "GetTeam should return non-nil for unassigned")
assert_equal(GetTeam(Team.kSpectator) ~= nil, true, "GetTeam should return non-nil for spec")

local red_team = GetTeam(Team.kRed)
local green_team = GetTeam(Team.kGreen)

assert_equal(red_team:GetPlayerLimit(), 0, "Red should have no player limit")
assert_equal(green_team:GetPlayerLimit(), -1, "Green should have -1 player limit")

assert_equal(red_team:GetClassLimit(Player.kScout), 0, "Red scout should have no class limit")
assert_equal(red_team:GetClassLimit(Player.kCivilian), -1, "Red civ should have -1 class limit")

red_team:ClearAllies()
assert_equal(#red_team:GetAllies(), 0, "Red should have no allies")
red_team:SetAllies(Team.kGreen)
assert_equal(#red_team:GetAllies(), 1, "Red should have 1 ally")
assert_equal(red_team:GetAllies()[1]:GetTeamId(), green_team:GetTeamId(), "Red should be allied with the green team")
red_team:ClearAllies()
assert_equal(#red_team:GetAllies(), 0, "Red should have no allies")
```

### DIFF
--- a/dlls/ff/ff_lualib_team.cpp
+++ b/dlls/ff/ff_lualib_team.cpp
@@ -39,19 +39,18 @@ namespace FFLib
 	luabind::adl::object GetAllies(CFFTeam *pTeam)
 	{
 		luabind::adl::object luatblAllies = luabind::newtable(_scriptman.GetLuaState());
-		if (pTeam)
+
+		int iTableKey = 1;
+		int alliesMask = pTeam->GetAllies();
+		for (int teamId = TEAM_UNASSIGNED; teamId < TEAM_COUNT; teamId++)
 		{
-			int iTableKey = 1;
-			int alliesMask = pTeam->GetAllies();
-			for (int teamId = TEAM_UNASSIGNED; teamId < TEAM_COUNT; teamId++)
-			{
-				if (alliesMask & (1<<teamId))
-				{
-					CFFTeam *pAlliedTeam = dynamic_cast<CFFTeam*>(g_Teams[teamId]);
-					luatblAllies[iTableKey++] = luabind::adl::object(_scriptman.GetLuaState(), pAlliedTeam);
-				}
-			}
+			if (!(alliesMask & (1<<teamId)))
+				continue;
+
+			CFFTeam *pAlliedTeam = dynamic_cast<CFFTeam*>(g_Teams[teamId]);
+			luatblAllies[iTableKey++] = luabind::adl::object(_scriptman.GetLuaState(), pAlliedTeam);
 		}
+
 		return luatblAllies;
 	}
 };
--- a/dlls/ff/ff_lualib_team.cpp
+++ b/dlls/ff/ff_lualib_team.cpp
@@ -8,6 +8,7 @@
 #include "cbase.h"
 #include "ff_lualib.h"
 #include "ff_team.h"
+#include "ff_scriptman.h"
 
 // Lua includes
 extern "C"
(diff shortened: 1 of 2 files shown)


dlls/ff/ff_lualib_team.cpp
```diff
@@ -32,6 +33,29 @@ std::ostream& operator<<(std::ostream& stream, const CTeam& team)
...
+	luabind::adl::object GetAllies(CFFTeam *pTeam)
+	{
+		luabind::adl::object luatblAllies = luabind::newtable(_scriptman.GetLuaState());
+		if (pTeam)
+		{
+			int iTableKey = 1;
+			int alliesMask = pTeam->GetAllies();
+			for (int teamId = TEAM_UNASSIGNED; teamId < TEAM_COUNT; teamId++)
+			{
+				if (alliesMask & (1<<teamId))
```
AfterShockFF: You can reduce nesting here by inverting the `if` and using `continue` (i prefer early-exit over nested if statements)

dlls/ff/ff_lualib_team.cpp
```diff
@@ -32,6 +33,29 @@ std::ostream& operator<<(std::ostream& stream, const CTeam& team)
 	return stream << "team" << ":" << const_cast<CTeam&>(team).GetName() << ":" << const_cast<CTeam&>(team).GetTeamNumber();
 }
 
+namespace FFLib
+{
+	// helper function to turn the bitmask into a lua table of teams
+	luabind::adl::object GetAllies(CFFTeam *pTeam)
+	{
+		luabind::adl::object luatblAllies = luabind::newtable(_scriptman.GetLuaState());
+		if (pTeam)
```
AfterShockFF: Unsue what luatblAllies is going to be at this stage but I think you can reduce nesting here too and just return immediately

```
if (!pTeam)
{
    return luatblAbllies;
}
```

dlls/ff/ff_lualib_team.cpp
```diff
@@ -62,8 +86,11 @@ void CFFLuaLib::InitTeam(lua_State* L)
 		class_<CFFTeam, CTeam>("Team")
 			.def("SetAllies",			&CFFTeam::SetEasyAllies)
 			.def("ClearAllies",			&CFFTeam::ClearAllies)
+			.def("GetAllies",			&FFLib::GetAllies)
```
AfterShockFF, line 88: I notice that the other functions here are tagged with CTeam or CFFTeam, but this is FFLib, whilst being under `class_<CFFTeam, CTeam>("Team")`, is this OK?

squeek502, line 88: A function that takes a class pointer as the first parameter is bound by Luabind the same as a member function of that class. `GlobalFunction(CFFPlayer *obj, int param)` is treated the same as if it were `CFFPlayer::Function(int param)`

From the [Luabind docs](http://www.rasterbar.com/products/luabind/docs.html#binding-classes-to-lua):

> It is also possible to register free functions as member functions. The requirement on the function is that it takes a pointer, const pointer, reference or const reference to the class type as the first parameter. The rest of the parameters are the ones that are visible in Lua, while the object pointer is given as the first parameter.

dlls/ff/ff_lualib_team.cpp
squeek502: Just realized that it's actually impossible for pTeam to be NULL here. Lua exposes the function as CFFTeam::GetAllies, so there's no way to call it with a NULL object.

PR #189, opened 2015-04-05
Fix explosions going through brushes when directly underneath them

squeek502:
- The z+1 seems to be carried over from HL1 code but is no longer necessary, and was causing things that explode underneath a brush to be pushed up into the brush and then hit things through the brush. Tested with a double bounced rail explosion (was 100% consistently reproducible).

Shock was right to be worried.


AfterShockFF:
Provided you already tested rockets and grenades etc exploding on the floor, seems fine :+1:

PR #190, opened 2015-04-05
Fix CFFPlayer noisy BodyTarget while crouched

squeek502:
- Was always returning a hardcoded view offset instead of taking crouching into account
- Should make explosions while crouched behave better

### DIFF
--- a/dlls/ff/ff_player.cpp
+++ b/dlls/ff/ff_player.cpp
@@ -7620,7 +7620,7 @@ Vector CFFPlayer::BodyTarget(const Vector &posSrc, bool bNoisy)
 
 	if (bNoisy)
 	{
-		return GetAbsOrigin() + (Vector(0, 0, 28) * random->RandomFloat(0.5f, 1.1f));
+		return GetAbsOrigin() + (GetViewOffset() * random->RandomFloat(0.5f, 1.1f));
 	}
 	else
 	{


dlls/ff/ff_player.cpp
```diff
@@ -7620,7 +7620,7 @@ Vector CFFPlayer::BodyTarget(const Vector &posSrc, bool bNoisy)
 
 	if (bNoisy)
 	{
-		return GetAbsOrigin() + (Vector(0, 0, 28) * random->RandomFloat(0.5f, 1.1f));
+		return GetAbsOrigin() + (GetViewOffset() * random->RandomFloat(0.5f, 1.1f));
```
AfterShockFF, line 7623: Whats this random bit for, out of interest?

squeek502, line 7623: That's what makes it 'noisy'. It chooses a random offset along the Z axis to test. 

Here's [HL1's for reference](https://github.com/ValveSoftware/halflife/blob/5c1e1d63eb5a9cb97b3b19f992d4e62aaae02cf0/dlls/player.h#L214) (note that FF's GetAbsOrigin returns the center of the player, so they will be functionally identical after this change):

``` c++
 virtual Vector BodyTarget( const Vector &posSrc ) { return Center( ) + pev->view_ofs * RANDOM_FLOAT( 0.5, 1.1 ); };
```

squeek502:
Merging as this is a straightforward bugfix.

Issue #188, opened 2015-04-05
What in the absolute world is going on with our explosion code?


squeek502:
What is this?
- https://github.com/fortressforever/fortressforever/blob/246e629aa25c9b2e497b45a46093419757a32e75/game_shared/ff/ff_gamerules.cpp#L1327-L1385

Is that still necessary? The last time it had any real changes was 2006.

It also uses 36 as 'half the model height' without checking if the player is crouched. That seems wrong. (EDIT: Should be fixed by #191)

For reference:
- Default HL2 RadiusDamage: https://github.com/fortressforever/fortressforever/blob/beta/game_shared/gamerules.cpp#L463-L470
- Default HL1 RadiusDamage: https://github.com/ValveSoftware/halflife/blob/5c1e1d63eb5a9cb97b3b19f992d4e62aaae02cf0/dmc/dlls/combat.cpp#L886-L958

DexterHaslem:
closing old issue that was cleaned up in PR

PR #191, opened 2015-04-05
Fix explosions doing too much damage to crouched players in certain cases

squeek502:
- Fixes the hitbox hack always using full height dimensions / ignoring the player being crouched
- Seems to affect things that explode above the player the most, but also affects things like crouched rocket jumping and railgun jumping
- See also #188


AfterShockFF:
Does this change how rocket jumps and stuff work? if so, in what way?

squeek502:
It only affects RJing/rail jumping while crouched, and only when you are a distance away from the explosion. It makes it so that the damage is actually consistent with your distance from the explosion; meaning, if you crouch and shoot a charged rail so that it explodes when you are at the peak of your jump, it will do less damage than if you shoot right as you jump. Before this change, that wasn't the case (it'd do max damage even though you weren't close enough to actually take max damage).

It's essentially a straight-forward bugfix, as it was using the wrong sized hitbox for crouched players.

AfterShockFF:
:+1:

PR #192, opened 2015-04-05
Lua: Add global ServerCommand function + some protection

squeek502:
- ServerCommand is a map-agnostic alternative to the point_servercommand entity
- Protect rcon_password from being changed by point_servercommand and SetConvar global Lua functions


AfterShockFF:
:+1:

PR #193, opened 2015-04-06
Fix HH nades not doing damage to nearby players while standing still

### LINKED ISSUE #24: Handheld grenades sometimes don't hurt nearby players
Somewhat sporadic but semi-reproducable by standing completely still and hand-helding a frag grenade with a player next to you. No clue what would be causing it.

squeek502:
- Not the best way to do it, but it seems to work
- Fixes #24


AfterShockFF:
:+1:

PR #194, opened 2015-04-06
Lua: Add player:GetArmorAbsorption

squeek502:
- Returns a float between 0 and 1 representing the percentage of damage that is absorbed by armor


dlls/ff/ff_player.cpp
```diff
@@ -5417,7 +5417,7 @@ int CFFPlayer::OnTakeDamage(const CTakeDamageInfo &inputInfo)
 		//float flNew = info.GetDamage() * flRatio;
 		float fFullDamage = info.GetDamage();
 
-		float fArmorDamage = fFullDamage * (((float)m_iArmorType) / 10.0f); //AfterShock: changing int to float e.g. armor type 8 means 0.8 i.e. 80% damage absorbed by armor
+		float fArmorDamage = fFullDamage * GetArmorAbsorption(); //AfterShock: changing int to float e.g. armor type 8 means 0.8 i.e. 80% damage absorbed by armor
```
AfterShockFF: Should probably move the comment down to the function now, since it's explaining that weird bit of code

AfterShockFF:
:+1:

Issue #50, opened 2014-11-09
Add armor-strip friendly fire support


squeek502:
See: http://forums.fortress-forever.com/project.php?issueid=444

How it works in TFC:

> The Mp_teamplay command sets up the teamplay settings for TFC, there are a number of settings that can be set in this, to make it easy I have listed some common ones which you can use.
> 
> Mp_teamplay 1 = TP is on (must be set)
> Mp_teamplay 11 = TP on with Friendly Firee
> Mp_teamplay 21 = TP is on standard setting, no FF health stripping but FF armor strip on.
> Mp_teamplay 1301 = TP is on, No friendly fire health or armor stripping from weapons or nades.
> Mp_teamplay 175381 = TP is on, Mirror Damage On.
> 
> These numbers are made up by adding together the specific mp_teamplay commands for example 1+4+16+256+1024=1301
> 
> 1 = Teamplay on (this should always be set)
> 2 = Teammates take half-damage from direct weaponfire
> 4 = Teammates take no damage from direct weaponfire
> 8 = Teammates take half-damage from explosive weaponfire
> 16 = Teammates take no damage from explosive weaponfire
> 128 = Teammates armor takes half-damage from direct weaponfire
> 256 = Teammates armor takes no damage from direct weaponfire
> 512 = Teammates armor takes half-damage from explosive weaponfire
> 1024 = Teammates armor takes no damage from explosive weaponfire
> 2048 = YOU take half damage from hitting Teammates with direct weaponfire
> 4096 = YOU take no damage from hitting Teammates with direct weaponfire
> 8192 = YOU take half damage from hitting Teammate with explosive weaponfire
> 16384 = YOU take no damage from hitting Teammate with explosive weaponfire
> 32768 = YOUR armor takes half damage from hitting Teammate with direct weaponfire
> 65536 = YOUR armor takes no damage from hitting Teammate with direct weaponfire
> 131072 = YOUR armor takes half damage from hitting Teammate with explosive weaponfire
> 262144 = YOUR armor takes no damage from hitting Teammate with explosive weaponfire

In FF, mp_teamplay does nothing iirc.

squeek502:
For future reference, in TFC, mp_teamplay 21 (full armor stripping) deals armor damage dependent on armor absorption, meaning that if a shot does 100 damage, and the teammate's armor absorbs 80%, then 80 armor damage would be done. The damage that would have been normally applied to health is discarded.

squeek502:
[...]

PR #195, opened 2015-04-06
Add basic armorstrip convar: mp_friendlyfire_armorstrip

squeek502:
From the cvar description:
  If > 0, only deals armor damage to teammates. The armor damage is multiplied by the value of the cvar (mp_friendlyfire_armorstrip 0.5 means half damage). Requires mp_friendlyfire 1

Contributes towards #50 (or closes it if we don't want to implement all of the features of mp_teamplay)

### DIFF
--- a/game_shared/ff/ff_gamerules.cpp
+++ b/game_shared/ff/ff_gamerules.cpp
@@ -181,10 +181,10 @@ ConVar mp_prematch( "mp_prematch",
 					FCVAR_NOTIFY|FCVAR_REPLICATED,
 					"delay before game starts" );
 
-ConVar mp_armorstrip( "mp_armorstrip",
+ConVar mp_friendlyfire_armorstrip( "mp_friendlyfire_armorstrip",
 					"0",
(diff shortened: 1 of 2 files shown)


game_shared/ff/ff_gamerules.cpp
```diff
@@ -181,6 +181,11 @@ ConVar mp_prematch( "mp_prematch",
 					FCVAR_NOTIFY|FCVAR_REPLICATED,
 					"delay before game starts" );
 
+ConVar mp_armorstrip( "mp_armorstrip",
```
AfterShockFF: I wonder if we should call this mp_friendlyfire_armorstrip? so when changing friendlyfire you get the intellisense and searching for friendlyfire brings this up.

AfterShockFF:
:+1:

game_shared/ff/ff_gamerules.cpp
squeek502: I like that idea.

PR #198, opened 2015-04-06
Revert "Fix schedule/timer/menuman memory leak"

squeek502:
Reverts fortressforever/fortressforever#173

> I'm going to revert this because I've been getting some mysterious crashes while changing map that seem to be related to Lua shutting down and freeing memory. Reverting this locally seemed to fix those crashes as well.
> 
> I guess I shouldn't try to fix what I'm not totally sure is broken.

Didn't know it'd open a new PR.


AfterShockFF:
Looks good :+1:

Issue #95, opened 2014-12-10
Allow customization without overwriting files


squeek502:
This is something the 2013 engine has, but I'm not totally sure if it's feasible for us. It'd be a really nice thing to have if we ever want to add workshop support, so that we don't have to step on game files to install things (which sounds like bad news).

From https://github.com/fortressforever/fortressforever/pull/88#discussion_r21584752:

> Did some more research about search paths. They are defined in FileSystem_Steam.dll and set based on [gameinfo.txt](https://github.com/fortressforever/fortressforever-scripts/blob/master/gameinfo.txt#L24-L47). Although we can't access the implementation in that dll, there is an analog (which seems to be accurate) in [public/filesystem_init.cpp](https://github.com/fortressforever/fortressforever/blob/beta/public/filesystem_init.cpp#L498-L590) that's used to compile demoinfo.exe.
> 
> Fortress Forever's current search paths:
> 
> ```
> NULL pathID search paths:
> c:\program files (x86)\steam\steamapps\common\fortress forever\platform\
> c:\program files (x86)\steam\steamapps\common\fortress forever\fortressforever\
> c:\program files (x86)\steam\steamapps\common\fortress forever\hl2\
> c:\program files (x86)\steam\steamapps\common\fortress forever\platform\config\
> 
> GAME search paths:
> c:\program files (x86)\steam\steamapps\common\fortress forever\fortressforever\
> c:\program files (x86)\steam\steamapps\common\fortress forever\hl2\
> 
> MOD search paths:
> c:\program files (x86)\steam\steamapps\common\fortress forever\fortressforever\
> 
> GAMEBIN search paths:
> c:\program files (x86)\steam\steamapps\common\fortress forever\fortressforever\bin\
> c:\program files (x86)\steam\steamapps\common\fortress forever\hl2\bin\
> 
> PLATFORM search paths:
> c:\program files (x86)\steam\steamapps\common\fortress forever\platform\
> 
> DEFAULT_WRITE_PATH search paths:
> c:\program files (x86)\steam\steamapps\common\fortress forever\fortressforever\
> ```
> 
[...]

PR #217, opened 2015-04-25
Make upgrading SGs require 3 consecutive clangs at full HP

### LINKED ISSUE #211: Give engi's a choice when upgrading SG's
> **squeek:**
> right, i play under the assumption that level 2 is good enough
> level 3 is a luxury
> **AfterShock:**
> yeah
> level 2 is a lot better than 1
> it didn't used to be that way
> **squeek:**
> which leads to the age old thing of 'i wish upgrading was a choice'
> **AfterShock:**
> but that doesnt mean this way isnt better
> (i mean level 3 used to be more common?)
> **squeek:**
> because getting a level 2 and then having 130 cells means your sg is as good as dead
> level 3 is just as common
> i think im the only one that will maintain level 2s
> but it's a bad idea because if you get >= 130 cells then your first repair costs 130 no matter what
> **AfterShock:**
> in the old days of tfc, you'd occasionally get big metal bags from soldiers who died with a lot of armor (after you repairing them)
> **squeek:**
> which is major lame
> **AfterShock:**
> well we could make it so you can only upgrade a repaired SG
> but whacking away would then sometimes accidenrally upgrade it
> **squeek:**
> yeah the ideal would be to make it a legit choice somehow
> but not hidden away as a menu option or something
> maybe a hold to upgrade thing
> where you do something different if you hold fire
> so to just repair you'd press/release over and over
> **AfterShock:**
> hmm
> **squeek:**
> we've always had good success with charging mechanics!
> not buggy at all!

squeek502:
- Allows players to intentionally avoid upgrading a sentry gun if they want to
- Closes #211

This is just one potential implementation to make upgrading a choice. It doesn't seem half bad. Hopefully it's not too much of a nerf.

To preview it with a sound, add this to game_sounds_ff_weapons.txt:

```
    "Spanner.PartialUpgradeSG"
    {
        "channel"       "CHAN_WEAPON"
        "volume"        "1.0"
        "wave"          "buttons/lever4.wav"
    }
```


squeek502:
Here's a quick video demonstration:

https://www.youtube.com/watch?v=4T-KN2x5GlQ

TYRANNOSAURUSCOW:
What about having "consecutive clangs" meaning no delay between clangs?

squeek502:
AfterShock and NeoNL both mentioned there should be a time threshold as well. I'll try adding that.

squeek502:
Here's an updated video with the timeout added and some potentially-final sounds.

https://www.youtube.com/watch?v=1Is7k1XGz_I

mikeparker:
I think:
* The cog sounds should be a bit louder
* The timeout should be shorter. I think if you stop clanging it should timeout near instantly. 
* I also think theres no need for a timeout sound.

AfterShockFF:
I think:
- The cog sounds should be a bit louder
- The timeout should be shorter. I think if you stop clanging it should timeout near instantly. 
- I also think theres no need for a timeout sound.

AfterShockFF:
After some discussion i think the timeout is probably ok as it is. Any repairs will stop the upgrade process and the longer timeout will help noobs maybe.

squeek502:
I want to try merging this. @FDA-, thoughts?

FDA-:
Uh, I have mixed feelings about it. I just feel like it could be a nuisance in some situations. Also you pointed out that you don't want this to be some option hidden away in a menu somewhere, but this implementation would be a completely undocumented feature, even with some kind of a hud notifier it has the potential to confuse some players. I keep feeling like over all it'd be a nuisance to engies, but no, it would actually be a decent buff to them. Having 200 cells offers you an effective 500% health? 100% base gun health, an extra 100% for every 50 cells.(ok obviously not quite that much, no one is going to be hitting each repair at 1-10% health, or have a full 200 cells all the time) In many situations I suppose it wouldn't really be a "choice" as it would be a buff to maintaining lvl 1 and 2 guns. I guess you have to decide if engies need/deserve that. I suppose engies actually already have this optionality, but it requires bag juggling, which has a potential down side of people picking up your bags. That's a more interesting interaction, but of course you're not going to see anyone do that. I also seem to remember discussion going around earlier about increasing build time. I just thought why in the world would you want to do that? Idk who started that, but to me it suggests someone out there thinks engies are already strong enough. If I can get in a FR and on top of the SG I can usually kill it, if I have to snipe it, a lot of the time it's a struggle, at least for a single medic. So seems like a direct buff to lvl 1-2 repairs, is that something you want, and is it something you're willing to basically force on a player? That's about what it comes down to imo.

squeek502:
I think you're right, but I'm not sure that's a bad thing. Something to keep in mind is that levels 1 and 2 die very quickly when not repaired, as each level has more health than the previous one (level 1 has 145, level 2 has 180, level 3 has 200). This creates a natural tradeoff to maintaining a level 1/2; the window of opportunity for saving the gun is smaller the lower level it is.

I think the choice aspect would only appear in certain situations, and it might allow struggling engis get a foothold when they are being constantly overrun. For example, if you're cell starved, you could stay on level 2 while getting up to 200 cells and then only upgrade once you know you'll have access to cells after the upgrade. Currently, you get stuck in a loop where it's hard to restabilize and going from level 2 to level 3 can be a constantly precarious position that you have to try over and over to get past (my guns die _extremely often_ just after I upgrade to level 3).

I wonder if a change like this could be introduced as a server variable at first or something, to get it tested without having to get people to test on a separate beta branch.

AfterShockFF:
I would also like to establish a way of testing new features, for example if we get around to testing spy stuff at some point. One way would simply be to push to the main branch for a day then roll it back after a day or so. You could also change the front end background to say "Beta testing" or something, to make it clear what's going on, and point people to a forum post to give feedback. 

Another would be to actually use a beta branch, which I think is supported by Steam by default, but I haven't looked into it. Steam has a "switch to beta testing branch" thing I think.

A 3rd method is like you said, to have a server variable. You could even have "sv_betatestingmode" or something which could be sent to the clients as well, so we could set something up to use longer term.

squeek502:
The trouble with using a Steam branch is that it requires people to run separate servers on the beta branch, which few people will want to do.

Putting things on a server variable would make it easier for users but a bit harder for developers.

AfterShockFF:
Is switching a server to the beta branch a difficult task? Is it much harder than tweaking a server variable? 

If we can use a server variable then that's easiest. I don't know if replacing classes wholesale might be trickier (like reading weapons/grenades from code rather than script files) but we can cross that bridge when we come to it.

squeek502:
Most people don't manage their servers themselves, so I'm not sure what tools server hosts have for that type of thing. It might be easy and it might be hard, but I'm not sure many people would do it either way.

A server variable would make things _really_ easy for server owners.

FDA-:
Pretty sure clients would have to opt into the test branch as well. Which makes opting in and out a hassle. That's the way it's been in other games I've played.

squeek502:
Right, and that could be a _legit_ hassle depending on how branch switching affects asset customizations, as we can't use the modern way of installing customs in a separate directory (see #95).

I think a server variable is the way to go.

WillWowmc:
Just checking in on this, what's the go with it? Anything happening in the end?

Issue #221, opened 2015-05-04
Get FF compiling on more recent Visual Studio


AfterShockFF:
Currently we can only develop on Visual Studio 2005 which is very old. From a quick test it seems like this is something to do with tier0 / tier1 valve libraries.

Upgrading would mean we can take advantage of a load of new features and stuff like Resharper C++.

DexterHaslem:
I've also been wanting to do this, and have tried in the past. another issue was luabind not building. newer version of luabind worked, but needed newer version of boost and I got stuck somewhere else I can't recall

DexterHaslem:
oh yea, only tier1 can be build from sources (in everything project), so screwed on tier0,  2,3 being built  on older versions (lib\public & lib-vc7\public) as far as i know. I'm not super familiar with the ABI changes over the yhears

a licensee could try to rebuild them but trying to build on newer version of VC++ would probably cause a million problems

AfterShockFF:
So it looks like we can't link to the old tier0/2/3 lib files with the newer compilers because they were compiled in ancient C++ and the compiler has had breaking changes since then, so we would either need to:
- Get hold of the source code and then update the source code to deal with the minor little things that have changed, then recompile in the newest compiler
- Keep compiling in VS2005 but somehow work out how to work in a newer VS and backport the project changes in order to compile (some automated project-change detection script?)
- Work out how to wire up a newer VS to use the old compiler, but still take advantage of all the newest VS features

AfterShockFF:
It looks like the 3rd option is probably possible - wire new visual studio up to use the old compiler. It's a bit of a pain to do, and I don't have it fully working yet, but you can use something called Daffodil:

https://daffodil.codeplex.com/

You need to install VS2010 then the latest VS afterwards.

http://stackoverflow.com/questions/15029762/is-it-possible-to-use-the-vc-6-compiler-in-visual-studio-2012/16487010

AfterShockFF:
Ok I can't get this to work - I'm sure someone else probably could if anyone wants to give it a try. The trick is to get it working in 2010 first, it seems. I got as far as making 2012 target the old 2005 compiler but it throws an error:

`C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\Platforms\Win32\Microsoft.Cpp.Win32.Targets(153,5): error MSB6006: "CL.exe" exited with code -1073741515.`

Which I can't figure out, there doesn't seem to be any useful feedback on exactly what's going wrong.

[...]

PR #229, opened 2015-06-03
Dev/modern vs projs

DexterHaslem:
works well for browsing the sources. still looking into a way to convert 2013->2005 so you could really go to town in 2013. per #221 actually compiling isn't really possible


DexterHaslem:
While you could technically browse w/ newer VS, VS code w/ the C++ extension does a fine job at it without the overhead of VS grinding away. toasting it

PR #245, opened 2015-08-11
Add sound to rampsliding

squeek502:
I'm pushing any WIP stuff I had sitting on my computer just in case anyone felt like picking it up where I left off. This should work, just need to get the right sound.


BullyFF:
https://soundcloud.com/tudberry/sets/ffsoundstuff
FFslide1 - 5 tracks ?

squeek502:
Right, I tested it with those sounds, but I think it needs to be more subtle. I don't really know what it should sound like, but it shouldn't be too distracting.

BullyFF:
Sound files now in my shared folder

Issue #252, opened 2016-05-10
weird positioning w/ hud_centerid 1


fryl0ch:
it looks like the hud_centerid 1 positioning code is [doing some wierd stuff with aspect-ratio scaling](https://github.com/fortressforever/fortressforever/blob/v2.46.1/cl_dll/ff/ff_hud_crosshairinfo.cpp#L592-594)

It looks centered in [4:3](http://i.imgur.com/c6e99U9.jpg), slightly off-center in [16:9](http://i.imgur.com/tB7Ro2t.jpg)... and on my setup [(48:9) the issue is a bit exaggerated](http://i.imgur.com/YjtejF3.png)

[imgur album](http://imgur.com/a/LuXqU)

fryl0ch:
looks fine in 16:9, 
![20160519192829_1](https://cloud.githubusercontent.com/assets/7390162/15412737/fa5cd788-1df7-11e6-9b5e-f43cd550c98d.jpg)

however, now in widescreen i'm not seeing anything
![20160519191308_1](https://cloud.githubusercontent.com/assets/7390162/15412756/2f8aca5a-1df8-11e6-9cea-713bfc126030.jpg)

Just got the build going, ill see what i can do with it

squeek502:
Interesting. I'd be curious to know what `surface()->GetScreenSize()` gives for your resolution. If you are comfortable with Visual Studio, set a breakpoint [here](https://github.com/fortressforever/fortressforever/blob/beta/cl_dll/ff/ff_hud_crosshairinfo.cpp#L586) (see [the readme for how to start a debugging session](https://github.com/fortressforever/fortressforever#debugging-an-active-session)) and see what `iScreenWide` and `iScreenTall` get set to.

The only real difference between this and what the crosshair is doing to set its position is that the crosshair uses `ScreenWidth()` while crosshairinfo is using `surface->GetScreenSize()`. I wonder if that's an issue.

fryl0ch:
![untitled-1 copy](https://cloud.githubusercontent.com/assets/7390162/15415405/706a8124-1e10-11e6-9b18-5f5627477fc3.jpg)

i think you might be on to something

squeek502:
Nice, appreciate you looking into this. The difference between `ScreenWidth()` and `surface()->GetScreenSize()` is that `ScreenWidth` gets the root panel size when it can (`ScreenWidth` calls `GetHudSize`):

```
void GetHudSize( int& w, int &h )
{
[...]

PR #255, opened 2016-05-20
Fix various HUD positioning on non-standard resolutions

squeek502:
- GetHudSize gets the root panel size when applicable which seems to be more reliable than the raw ScreenSize
- See #252


squeek502:
@fryl0ch, let me know if this fixes your issue.

squeek502:
Just gonna merge this; still unconfirmed that it'll fix issues with giant resolutions, but the changes should be fine regardless.

PR #257, opened 2016-05-28
Cheaper dispenser

AfterShockFF:
- Reduced dispenser cost from 100 to 30 cells
- Reduced dispenser max ammo capacity from 400 to 100 cells
- Increased dispenser restock speed from 40 cells every 10 seconds to 20 cells every 2 seconds (2.5x faster)
- Reduce dispenser health from 150 to 75

Why?
- To make the dispenser usable when the engineers team is under a lot of pressure
- To reduce the effectiveness when the engineers team is doing well
- To reduce the reliance on using the railgun to recharge cells (so he can fight inbetween more)
- To increase the possibilities for using the dispenser as an explosive
- To reduce the time spent destroying dispensers

Basically, the engineer should be a building class, tinkering around with things and managing his cell stock. Currently, dispensers are quite expensive and often if the game is frantic, the engineer does not have the spare cells to build a dispenser - the priority is often to get the SG up as soon as possible. This is exacerbated if the map does not contain any nearby safe dispenser locations. 

A while ago, we normalised death bags so they always contain 30 cells, including engi death bags, which could be up to 200 metal, similarly with defensive players such as soldiers dying


FDA-:
I'd like to take a moment to talk about the greater implications of the beta branch, as this pull has sort of popped up with a lot of that other stuff.

First for this dispenser change in specific, I do find it interesting. I was kind of confused about all the strange value changes. But I do like the logic going into this:

Strengthen dispenser under stress; weaken dispenser in clout.

I'm not really sure the dispenser really needs to be weaker when the defense is strong. I've definitely experienced the opposite end of how strong an ever-present dispenser can be, taking down a sentry just to see a new one go up to level three in no time. But I'm not really sure that means it needs to be weaker. Just because all in all, it's not that hard to take down if it's actually targeted, and if you take the time to target it, you'll likely take it down regardless of this health differential.

I think it's one of those things that just comes down to the community driven meta really. If people really concentrated on dispensers like they should, you'd see them go down all the time. It's just not one of those things you really see happen at this lower level of competitive play we keep up these days.

That being said I don't think lowering dispenser HP would ruin anything, but 75 might be drastic. Maybe 100, or 90 would be better.

Anyways, on to the broader beta branch. I see a lot of stuff floating around in this git feed about some of the changes I've see in the old, outdated beta. Like all the spy changes, the new engi gren.

When you guys kind of decided, "why not put these movement weapons in? (grapple hook and jump gun) What's the worst that could happen?" I was pretty much in agreement. Those are big, interesting changes, but at the same time they really don't stray too far from the FF we know and love. These other changes, like the spy changes, and even the engi secondary gren (as insignificant as that one is) kind of worry me. These are the type of changes that could really upset our fragile little community. The movement tools I think are fine to go in and get a "pseudo open-beta test", because our community is sort of built around movement. It will be a pretty natural and comfortable process for us to gain these tools and find out how they play into the rest of FF. These spy changes though, I do not think it's responsible or at all advisable to implement without some thorough testing, as in real, testing, not just dumping on the public.

The biggest complaint about spy really, is that the class isn't fun at all to play against, and there's always the chance that these changes might not make it any more fun to play against, and could completely ruin the experience for people who love playing the class now.

The same goes for the engi secondary nade. It just doesn't make sense  to change it now. (and to change it to something so static and so dull) You can throw all the game logic you want at me, about how the emp doesn't really fit the class, yada blah. But it's a fun grenade for the engi, and it's not a huge negative to those playing against the engi, so removing it just seems like a negative to me.

Especially when you're just going to replace it with a watered down nail grenade, after your holy crusade against nail grenades...

And just please, never speak of, or let see the light of day, those abominations that are the demo changes. Thank you.

squeek502:
Did someone say magnetic pipes? :rocket: :guitar:

For what it's worth, I mostly agree with FDA (especially about the hoverturret gren; I was never really sold on that). I don't have a strong opinion either way about these disp changes, though. I could never really get a feel for how they affected things when we tested them in the beta way back when.

AfterShockFF:
Thanks for the long response - appreciate the effort there. Lets catch up in IRC because theres too much to cover here.

PR #259, opened 2016-05-31
Use cl_gib_lifetime for dropped weapons from dead players

AfterShockFF:
and some minor refactoring


squeek502:
Fair enough. :+1:

PR #260, opened 2016-05-31
Gib more often when killed by explosions (-30hp compared to -50)

### LINKED ISSUE #184: Dynamic gib threshold.
As we all know, the biggest issue with FF has always been about how cool gibs look. As it stands at the current threshold, a player at 1 health can be gibbed by the SSG, yet so many explosives rarely gib still. I think the whole 1hp shotgun gib is neat and all, but to change the threshold any further would make even more shotgun gibs which would just start getting goofy.

So maybe we could have different thresholds for different weapons? I'd honestly like to see explosives gib a lot more often, 50% of the time if not more. The occasional ragdoll is cool, but bathing among the innards of your enemy is waaay more cool.

AfterShockFF:
Closes #184

### DIFF
--- a/dlls/ff/ff_player.cpp
+++ b/dlls/ff/ff_player.cpp
@@ -139,6 +139,9 @@ int g_iLimbs[CLASS_CIVILIAN + 1][5] = { { 0 } };
 ConVar ffdev_gibdamage("ffdev_gibdamage", "50", FCVAR_FF_FFDEV_REPLICATED, "If a players health is -(ffdev_gibdamage's value) or less after death, then they will gib instead of ragdoll");
 #define FFDEV_GIBDAMAGE ffdev_gibdamage.GetFloat()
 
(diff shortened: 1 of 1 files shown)


FDA-:
+1

squeek502:
Nice solution to avoid shotgun gibs. :+1:

FDA-:
Uh, except shotgun gibs are pretty much the best. Besides I think it only happens when you meat shot someone who is literally one health. Seems like it anyways. It's like a fatality, or the cherry on a DM cake. Doesn't happen all the time but when it does you know you messed that guy up. Or you know get rid of them w/e.

squeek502:
This doesn't change that--it makes the threshold for explosions lower without changing the threshold for non-explosions. So I meant 'good way to make gibs more common without also making shotgun gibs more common.'

FDA-:
Oh. +2

PR #261, opened 2016-05-31
Reduce default gib lifetime from 10 sec to 4

AfterShockFF:


FDA-:
Will this really make a big impact?

squeek502:
No opinion on this.

Note, though, that any current players won't be affected at all--defaults for existing `FCVAR_ARCHIVE` cvars only take effect when config.cfg doesn't exist.

PR #262, opened 2016-05-31
Reduce ragdoll lifetime from 15 to 5 sec

AfterShockFF:


squeek502:
It'd be nice if we could get this to be able to be controlled client-side, but doing so is harder than it might seem, unfortunately (I tried a while back and the mess of server-client interaction caused CL_CopyExistingEntity errors). 

No real opinion on this--I know you have a crusade against ragdolls. :smile:

PR #264, opened 2016-06-04
WIP: Sniper laser always visible

AfterShockFF:
Always show sniper laser, also make it originate from the gun rather than the snipers eyes.

Work in progress to gather ideas. 

Basically this means that whilst the sniper is charging the rifle, his laser is visible at all angles. It means that team-mates of the person being sniped can also see the laser as well as the person about to die. It also encourages snipers to hide a little bit more and be more sneaky rather than just holding down zoom and waiting for enemies.

Other options could include having a tracer bullet fire so team-mates see where a shot has come from momentarily.

### DIFF
--- a/game_shared/ff/ff_weapon_sniperrifle.cpp
+++ b/game_shared/ff/ff_weapon_sniperrifle.cpp
@@ -16,7 +16,8 @@
 static int g_iBeam, g_iHalo;
 
 #ifdef CLIENT_DLL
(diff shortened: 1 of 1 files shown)


Jitterskull:
Why we can't just have a sniper the way he was in 2.0 version of FF?? IMHO he was much more balanced back then. Well, except maybe keep the damage of AR the same way it's now.

PR #266, opened 2016-06-05
Add viewmodel charge effects to the jump gun

squeek502:
- Basically the railgun effects, but red instead of blue, and with a more pronounced effect at 100% charge


squeek502:
:+1:

PR #271, opened 2016-06-05
Increase railgun glow when fully charged

AfterShockFF:
This mirrors the change to the jumpgun code squeek made yesterday


squeek502:
Cool with me. :+1:

PR #274, opened 2016-06-05
Fix weapon-specific crosshairs for tranq/ac/rpg

squeek502:
See http://forums.fortress-forever.com/showpost.php?p=510354&postcount=2


AfterShockFF:
:+1:

PR #276, opened 2016-06-11
Refactoring: Create ff_player_shared.h file

AfterShockFF:
For ages we've had duplicate code across ff_player and c_ff_player, it's dumb and i've finally got around to cleaning some of it up. 

I'll add more to this PR as I find more shared stuff.


AfterShockFF:
Ok this was way harder than i thought. I might have to abandon this work.

PR #277, opened 2016-06-17
Set default scout spawnweapon to jumpgun

squeek502:
Not sure what's up with the change to the top of the file, made the change through the Github web UI. Probably a line ending thing?


AfterShockFF:
:+1:

PR #282, opened 2016-06-18
Improve Lua angle/vectors a bit

squeek502:
- Add player:GetEyeAngles()
- Add __tostring() implementations for Vector and QAngle classes
- Convert VectorAngles and AngleVectors to use Lua-style return values rather than C++ style pass-by-reference parameters. Example usage of the new functions:
  
  local eye_angles = player:GetEyeAngles()
  local forward, right, up = AngleVectors(eye_angles)
  local angles = VectorAngles(forward)
  assert(angles == eye_angles)

Note: This change is NOT backwards compatible, but as far as I can tell, no existing Lua scripts use either VectorAngles or AngleVectors currently (at least none that I have downloaded on my computer)

### DIFF
--- a/dlls/ff/ff_lualib_math.cpp
+++ b/dlls/ff/ff_lualib_math.cpp
@@ -26,12 +26,35 @@ extern "C"
 //---------------------------------------------------------------------------
 using namespace luabind;
 
-// Valve, you fucking suck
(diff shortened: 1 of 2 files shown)


AfterShockFF:
Seems fine, whats this for?

squeek502:
No specific purpose, just something I noticed could be improved when testing other changes.

Issue #250, opened 2016-04-18
Update or remove hints


AfterShockFF:
We probably have lots of hints that are outdated now, such as mentions of scout caltrops, nail grenades etc.

squeek502:
The state of this is actually not too bad. NeoNL also gave us a set of updated hints a while back on the beta forums that I'm not sure we ever did anything with:

http://forums.fortress-forever.com/showthread.php?t=23242

doublah:
Can this be closed? (From #285)

PR #285, opened 2016-06-19
Add new hints NeoNL wrote ages ago

alaswell:
Requires fortressforever/fortressforever-scripts#57
See fortressforever/fortressforever#250


squeek502:
Nice one. :+1:

PR #286, opened 2016-06-19
Fix performance issues with SpriteTrails when their parent teleports

### LINKED ISSUE #280: Flag trails flick across the map when capturing the flag
I get a bit of a stutter on capture too, but that could be unrelated. (might be flaginfo icons updating or something)

squeek502:
- Fixes stutter when the flag returns after it is capped (closes #280)
- Adds cvar cl_spritetrail_maxlength that controls the max length of a sprite trail (once exceeded, the trail will start over from its parent's current origin)
- Makes it so trails are not calculated at all client-side if they are disabled by their respective cvars


AfterShockFF:
:+1:

PR #287, opened 2016-07-03
Feature/kill assists

DexterHaslem:
ready to start soliciting feedback for basic kill assist support. once we know it's working
i'm all for making a nicer death notice and clarifying the nightmare of friendly-fire cases


DexterHaslem:
gunna create a PR from fortressforever feature/kill-assists directly so its easier for everyone! yay!

PR #288, opened 2016-07-03
Feature/kill assists

DexterHaslem:
soliciting feedback on first pass kill assist. things that have come up:

check damage coming from buildables like sentry, det explosion, etc. - consider GetDeathScorer if needed

death notice alignment is completely screwed up


DexterHaslem:
See also fortressforever/fortressforever-scripts#59 for required event script updates

dlls/ff/ff_player.cpp
```diff
@@ -8009,3 +8030,132 @@ void CFFPlayer::UpdateCamera( bool bUnassigned )
 		}				
 	}
 }
+
+// purpose: update our vector of recent attackers for kill assists
+void CFFPlayer::AddRecentAttacker( const CTakeDamageInfo &dmgInfo )
+{
+	CFFPlayer *pAttacker = ToFFPlayer( dmgInfo.GetAttacker() );
```
squeek502: Just noting that this is where `GetDeathScorer` could/should be used:

``` c++
CBasePlayer *pAttacker = g_pGameRules->GetDeathScorer(dmgInfo.GetAttacker(), dmgInfo.GetInflictor());
```

dlls/ff/ff_player.cpp
```diff
@@ -8034,9 +8034,13 @@ void CFFPlayer::UpdateCamera( bool bUnassigned )
 // purpose: update our vector of recent attackers for kill assists
 void CFFPlayer::AddRecentAttacker( const CTakeDamageInfo &dmgInfo )
 {
-	CFFPlayer *pAttacker = ToFFPlayer( dmgInfo.GetAttacker() );
+	// make sure to get the effective scorer incase dmg is from
+	// a buildable - det, sg, disp, etc
+	CFFPlayer *pAttacker = ToFFPlayer(dynamic_cast<CMultiplayRules *>(g_pGameRules)->GetDeathScorer( dmgInfo.GetAttacker(), dmgInfo.GetInflictor() ));
```
squeek502: This `CMultiplayRules*` cast should be unnecessary.

squeek502:
Looks like logging assists to file is not implemented yet. Here's our discussion about it from a while back:

```
Ryan Liptak:    do assists get logged as well?
Dexter Haslem:  in console? yea i just added to end of the other messages
Dexter Haslem:  (assisted by %s). etc
Ryan Liptak:    i mean server-side
Ryan Liptak:    might be two places
Dexter Haslem:  i think so, it does some message right that gets logged
Dexter Haslem:  i tried to look for the added airshot mod
Dexter Haslem:  might have missed a spot, i will chekc
Ryan Liptak:    https://github.com/fortressforever/fortressforever/blob/29d77663fc6512158f9f683970da3e1a6445f9e0/dlls/EventLog.cpp#L165
Dexter Haslem:  yea i missed that
Dexter Haslem:  hard
Dexter Haslem:  was going to add to end w/ a space between to not break existing parsing
Ryan Liptak:    not sure if it'd be better on the same line or as a separate line
Dexter Haslem:  oh yea thats a good idea for logs actually
Ryan Liptak:    not sure what a standalone line should look like either though
Dexter Haslem:  eh ill fuck with it
Dexter Haslem:  de facto answer
Ryan Liptak:    because the weapon isnt relevant
Ryan Liptak:    for the assister
Dexter Haslem:  yea totally not even tracked as well
Dexter Haslem:  just raw dmg
Ryan Liptak:    i guess just something like "\"%s<%i><%s><%s>\" assisted in killing \"%s<%i><%s><%s>\"
Ryan Liptak:    or assisted with
Dexter Haslem:  %s %s %s %s %s %s %s %s %s %s %s %s %s %s %s %s %s %s
Dexter Haslem:  yea thats prob good enough
Ryan Liptak:    maybe output the damage they did in the assist if that's accessible
Dexter Haslem:  dunno, cuz i dont clear on healing, and assisters that havent hit someone in the last 5 seconds ignored
Dexter Haslem:  pulled numbers out of my butt
Ryan Liptak:    right
Dexter Haslem:  would damage be useful at all tho? its pre-armor scaling and everything
Dexter Haslem:  seeing a soldier take 250 dmg might be confusing
Ryan Liptak:    who knows, but if we have access to it, why not throw it in there
Dexter Haslem:  then again this is *logs*
Dexter Haslem:  yea
Ryan Liptak:    only for the assist log
Dexter Haslem:  ok
Ryan Liptak:    keep the kill log line the same
```

dlls/ff/ff_player.cpp
DexterHaslem: It absolutely is unneeded. I made the unfortunate and often fatal mistake of randomly copying it out of some other FF code

DexterHaslem:
oops. I forgot entirely about that logging obviously. Assists on new line? would changing the kill line break any existing parsers

squeek502:
> would changing the kill line break any existing parsers

Yes, almost certainly. Assists on their own line probably isn't ideal, but at least it'd be backwards compatible. Note that, if assists are on their own line, it'd still be possible to associate assists with specific kills, because it'll be guaranteed that an assist is related to the kill line directly before it.

dlls/ff/ff_player.cpp
DexterHaslem: actually,  I was surprised, downcast is needed here
`error C2039: 'GetDeathScorer' : is not a member of 'CGameRules'`

squeek502: Ah, using `FFGameRules()` instead of `g_pGameRules` might be better then.

squeek502:
@DexterHaslem nice work. I'm unsure if the assist line should be made more stand-alone or not, though (e.g. `"%s<%i><%s><%s> assisted in killing %s<%i><%s><%s>"`).

DexterHaslem:
OK, I matched the kill line formatting (in quotes) too

squeek502:
Perfect. This good to go, then?

@AfterShockFF, any comments?

dlls/ff/ff_player.cpp
```diff
@@ -831,6 +835,10 @@ void CFFPlayer::PostThink()
 		m_angEyeAngles = EyeAngles();
 
 		m_PlayerAnimState->Update( m_angEyeAngles[YAW], m_angEyeAngles[PITCH] );
+
+		// check kill assists, drop off old ones
+		// this is actually done when queried when a player dies instead of every frame
+		//UpdateRecentAttackers( );
```
AfterShockFF: Looks like you can delete this commented code

dlls/ff/ff_player.cpp
```diff
@@ -8009,3 +8030,135 @@ void CFFPlayer::UpdateCamera( bool bUnassigned )
...
+			m_recentAttackers[i].totalDamage += dmg;
+			m_recentAttackers[i].timestamp = timestamp;
+			DevMsg( "CFFPlayer::AddRecentAttacker1: totalDamage = %f\n", m_recentAttackers[i].totalDamage );
+			return;
+		}
+	}
+
+	// if we didnt find a match, create & add new
+	DevMsg( "CFFPlayer::AddRecentAttacker2\n" );
+	//m_recentAttackers.AddToHead( RecentAttackerInfo( attackerIdx, dmg, timestamp, pAttacker ) );
```
AfterShockFF: I guess you can remove this

dlls/ff/ff_player.cpp
```diff
@@ -8009,3 +8030,135 @@ void CFFPlayer::UpdateCamera( bool bUnassigned )
...
+			continue;
+
+		// if this was a suicide and we are killerToIgnore, we dont have to check here
+		// because previous logic prevents adding ourselves to the assist list
+		bool isKiller = !killedByWorld && pFFAssister == killerToIgnore;
+
+#ifdef _DEBUG
+		DevMsg( "CFFPlayer::GetTopKillAssister1: assister='%s' cum dmg='%f' isKiller='%s'\n",
+			pFFAssister->GetPlayerName( ), m_recentAttackers[i].totalDamage, isKiller ? "yes" : "no" );
+#endif
```
AfterShockFF: If all this works then I would remove the dev messages, just clutters the code and the console for the next guy

dlls/ff/ff_player.cpp
```diff
@@ -8009,3 +8030,135 @@ void CFFPlayer::UpdateCamera( bool bUnassigned )
...
+void CFFPlayer::RemoveMeFromKillAssists( ) 
+{
+	// im disconnecting or switching teams or something, so 
+	// remove my contributions from assists. we can revisit this 
+	// if its too aggressive for eg, team switch and then having assist come through friendly fire
+
+	// if coming from spec or whatever, dont do anything because we couldnt have possibly damaged anything
+	// we are called before team is changed, so we don't have to worry about switching TO spec
+	if( GetTeamNumber() < TEAM_BLUE )
+		return;
```
AfterShockFF: What happens if a player shoots someone in the same frame  as going spec?

dlls/ff/ff_player.cpp
```diff
@@ -8009,3 +8030,135 @@ void CFFPlayer::UpdateCamera( bool bUnassigned )
...
+#ifdef _DEBUG
+			DevMsg( "CFFPlayer::GetTopKillAssister2" );
+#endif
+			continue;
+		}
+
+		if ( !ret || m_recentAttackers[i].totalDamage > ret->totalDamage )
+		{	
+			ret = &m_recentAttackers[i];
+			// consider in case of exact tie, take newest timestamp attacker here
```
AfterShockFF, line 8101: I was thinking the same thing but it's not a big deal

AfterShockFF:
One last point - how much of the screen is taken up if people have really long names?

cl_dll/ff/ff_hud_deathnotice.cpp
```diff
@@ -264,10 +265,9 @@ void CHudDeathNotice::Paint()
 		}
 		else
 		{
-
-			wchar_t victim[ 256 ];
-			wchar_t killer[ 256 ];
-
+			wchar_t victim[ 256 ];			
+			wchar_t killer[ 512 ]; // bumped this up for kill-assist msgs
```
AfterShockFF: we're not actually going to show all these characters on screen, are we?

squeek502: Yeah, might be overkill. See [MAX_PLAYER_NAME_LENGTH](https://github.com/fortressforever/fortressforever/blob/9e3d589b33c821f16627607d76b63b27d2844caf/public/const.h#L36) which is set to 32 (effectively 31).

dlls/ff/ff_player.cpp
AfterShockFF: (or more likely, their grenade / rocket explodes the same frame as going spec)

squeek502: Not actually sure this code could protect against that, as if you deal damage as you change to spec, then that damage would be dealt after `RemoveMeFromKillAssists` is run. Specs getting assists would need to be protected against in `GetTopKillAssister` instead (adding a `GetTeam` validation check).

That said, I'm not sure it hurts to run RemoveMeFromKillAssists regardless of current team?

AfterShockFF: Yeah it was more of a general comment to ensure we dont get random crashes due to all this code in general rather than this one block.

dlls/ff/ff_player.cpp
DexterHaslem: What in particular, the dev msgs? this is the guts of assist really

cl_dll/ff/ff_hud_deathnotice.cpp
DexterHaslem: the ui thingies measure and draw only used width but i'll change the size back

squeek502: Yeah the 256 thing is just something typical of FF code--when in doubt, make large capacity strings. It could probably be brought down a lot, but I'm not sure how much it matters.

dlls/ff/ff_player.cpp
squeek502: I would assume the commented line.

Jitterskull:
Would love so much the kill-assist feature for Fortress Forever. That's weird that the game that pretty much relies on team work and (at least used to have) different class cooperation between each other, didn't get it yet since 2007, especially considering that pretty much almost every single HL2 mod, even on 2004's source engine have kill-assist feature.

This would make fair share of points for doing damage and doing more effort rather than be the last person who do burst damage weapon and just kill someone with shotgun with one shot across the map without even doing much of effort, would also enourage less these assholes who do frag steal and finish someone off with instakilling weapon when you put so much effort in to killing someone and left enemy with just 5hp. It would as well make cheap suicide in combat far less forgiving and be used as easy exploit trick way to avoid giving enemy a point (especially for DM gamemode), which is a great thing indeed.

On some servers, you get exp for doing damage. Something that I would love a lot in a game like this.

squeek502:
So this is good to go now, right? Do we want to add an assists column to the scoreboard?

DexterHaslem:
Good idea I'll add that

DexterHaslem:
ignore build server results, I just turned it off since its impossible to build outside IDE :frowning:

PR #289, opened 2016-07-03
Fix local weapon sounds (like deploy sounds) playing twice on high ping

squeek502:
- No need to send the sound event from the server when the sound is only going to be heard on the client anyway
- Also made the sound play for observers if WeaponSoundLocal is called on their client, too (not a guarantee)
  - Could potentially be improved for observers by playing the sound on the client for the local player and sending the sound to observers on the server

See https://github.com/fortressforever/fortressforever/commit/1fd986d23bf9230920ff34b73f7eb9f7137fe688#commitcomment-18106476

### DIFF
--- a/game_shared/ff/ff_weapon_base.cpp
+++ b/game_shared/ff/ff_weapon_base.cpp
@@ -179,6 +179,7 @@ void CFFWeaponBase::WeaponSound(WeaponSound_t sound_type, float soundtime /* = 0
 //----------------------------------------------------------------------------
 void CFFWeaponBase::WeaponSoundLocal( WeaponSound_t sound_type, float soundtime )
 {
+#ifdef CLIENT_DLL 
 	// If we have some sounds from the weapon classname.txt file, play a random one of them
 	const char *shootsound = GetWpnData().aShootSounds[ sound_type ];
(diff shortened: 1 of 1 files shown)


AfterShockFF:
Would it be simpler to use this https://github.com/fortressforever/fortressforever/blob/beta/game_shared/SoundEmitterSystem.cpp#L1175-1199 ?

I have a feeling thats why I wrote that bit in the middle.

squeek502:
Not really the same thing--WeaponSoundLocal is meant to _only_ be heard by the local player (viewmodel sounds), so using `CSingleUserRecipientFilter` makes more sense. There's no reason for any server-side stuff except potentially for observers (but that wouldn't have worked before anyway).

AfterShockFF:
Ok so the server side sound you removed from WeaponSoundLocal was a bug, and won't have any side effects?

squeek502:
Correct. See the comment above the function:

```
//----------------------------------------------------------------------------
// Purpose: For weapon sounds that only the local player should hear
//----------------------------------------------------------------------------
```

Before, the server was still only sending it to the owner, which serves zero purpose except doubling the sound on high ping.

AfterShockFF:
OK :+1:

PR #290, opened 2016-07-07
Fix certain fields not getting saved in SourceTV demos

squeek502:
This will fix all the following fields not getting saved in SourceTV demos (+ pyro jetpack fuel once that's added):

```
BEGIN_SEND_TABLE_NOBASE( CFFPlayer, DT_FFPlayerObserver )
    SendPropFloat(SENDINFO(m_flNextClassSpecificSkill)),
    SendPropFloat(SENDINFO(m_flTrueAimTime)),
    SendPropFloat(SENDINFO(m_flHitTime)),
    SendPropInt(SENDINFO(m_nButtons)),
END_SEND_TABLE()
```

The one potential caveat here is that `m_nButtons` changes _a lot_, so saving it in the SourceTV demo for each player might inflate the size of demos considerably. The `m_nButtons` field is only needed for `hud_keystate`, so if we think it's too risky we could intentionally remove it from getting sent to SourceTV.

### DIFF
--- a/cl_dll/ff/ff_hud_keystate.cpp
+++ b/cl_dll/ff/ff_hud_keystate.cpp
@@ -225,11 +225,6 @@ bool CHudKeyState::ShouldDraw()
 		return false;
 
 	C_FFPlayer *pLocalPlayer = C_FFPlayer::GetLocalFFPlayer();
-
-	// this doesn't work in sourcetv demos, so don't even try
-	if ( engine->IsPlayingDemo() && engine->IsHLTV() )
-		return false;
-
 	C_FFPlayer *pTarget = C_FFPlayer::GetLocalFFPlayerOrAnyObserverTarget();
 
 	if( !pTarget ) 
--- a/dlls/ff/ff_player.cpp
+++ b/dlls/ff/ff_player.cpp
@@ -355,10 +355,10 @@ LINK_ENTITY_TO_CLASS( player, CFFPlayer );
 PRECACHE_REGISTER(player);
 
 BEGIN_SEND_TABLE_NOBASE( CFFPlayer, DT_FFPlayerObserver )
-	SendPropFloat(SENDINFO(m_flNextClassSpecificSkill)),
+	SendPropTime(SENDINFO(m_flNextClassSpecificSkill)),
 	SendPropFloat( SENDINFO( m_flJetpackFuel )),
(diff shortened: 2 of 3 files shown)


AfterShockFF:
Whats m_flNextClassSpecificSkill needed for, is that to display the overpressure hud bar? I wonder if we could round that to 8 bits rather than 32, since 256 segments on screen is probably fine. Similarly can we squeeze flHitTime down at all?

I know it seems like potentially over optimising but if it saves everyone disk space, download time and provides a good standard for future I think its worth it.

squeek502:
Yes, `m_flNextClassSpecificSkill` is for overpressure (the cooldown also shows on the AC viewmodel).

You're right that those floats seem like they can be optimized. Here's some potentially similar fields; I couldn't actually find many examples of `curtime` fields:

```
// playerlocaldata.cpp
    SendPropFloat   (SENDINFO(m_flDucktime), 12, SPROP_ROUNDDOWN|SPROP_CHANGES_OFTEN, 0.0f, 2048.0f ),
    SendPropFloat   (SENDINFO(m_flDuckJumpTime), 12, SPROP_ROUNDDOWN, 0.0f, 2048.0f ),
    SendPropFloat   (SENDINFO(m_flJumpTime), 12, SPROP_ROUNDDOWN, 0.0f, 2048.0f ),
```

```
// player.cpp
        SendPropFloat       ( SENDINFO( m_flDeathTime ), 0, SPROP_NOSCALE ),
```

I'm not really sure what the limits of `curtime` are.

squeek502:
Just realized that there's also a `SendPropTime` function:

``` c++
// basecombatweapon_shared.cpp
    SendPropTime( SENDINFO( m_flNextPrimaryAttack ) ),
    SendPropTime( SENDINFO( m_flNextSecondaryAttack ) ),
```

What it looks like (I'm not really sure what -1 bits does or SPROP_NOSCALE):

``` c++
SendProp SendPropTime(
    char *pVarName,
    int offset,
    int sizeofVar )
{
//  return SendPropInt( pVarName, offset, sizeofVar, TIME_BITS, 0, SendProxy_Time );
    // FIXME:  Re-enable above when it doesn't cause lots of deltas
    return SendPropFloat( pVarName, offset, sizeofVar, -1, SPROP_NOSCALE );
}
```

AfterShockFF:
Is this ready to merge?

squeek502:
Yes

Issue #3, opened 2014-10-15
Explosion falloff not related to radius


squeek502:
See: 
- https://github.com/fortressforever/fortressforever/blob/f663579932d1a685a2cb9195d3c25c3393c29b08/game_shared/ff/ff_gamerules.cpp#L1286
- https://github.com/fortressforever/fortressforever/blob/f663579932d1a685a2cb9195d3c25c3393c29b08/game_shared/ff/ff_gamerules.cpp#L1418

This makes falloff unreliable, and makes things that have an altered damage:radius ratio (the IC and frag grens most notably) either do zero damage before getting to the edge of its radius (in the case of the IC) or do more than 0.5x damage at the edge of its radius (in the case of frag grens).

Note: There would be knock-on effects to gameplay if this was fixed, and should only be fixed if we have a plan in place to deal with those.

Here's a commit that reverted some accidentally committed test code that implements a fix and prints the differences: 24eb0aa5e59ea64ee7b52564113551cbd34c4328

squeek502:
I misremembered the effects of this. Turns out most explosions do less than they would if the falloff was "fixed" (meaning 0.5x damage at the edge of the explosion for all explosions).

Damages close to the edge of the explosion radius:

| Thing exploding | Base damage | "Fixed" damage | Current damage |
| --- | --- | --- | --- |
| Frag Grenade | 145.000000 | 68.820503 | 3.148529 |
| IC Rocket | 55.000000 | 26.769991 | -9.159111 |
| RPG Rocket | 102.000000 | 49.961639 | 50.981998 |
| Grellow Pipe | 108.000000 | 50.931957 | 39.306984 |
| MIRV | 180.000000 | 74.475624 | 21.713440 |
| MIRVlet | 145.000000 | 57.599770 | -17.745255 |
| Unbounced fully charged rail | 40.000000 | 20.641932 | -8.395172 |
| Backpack by an EMP | 50.000000 | 33.925419 | -27.157990 |

So... yeah. It has wide-reaching and substantial effects (except on RPG rockets).

mikeparker:
Good findings, nice table.

I think personally, that I agree with valve's initial direction - that if you get hit by an explosion, you should know that you've taken at least a reasonable amount of damage. I think the game is more consistent and easier to read if you don't get hit for 1hp by a mirvlet, or 0hp from an IC rocket!

I think there are two variables we need to decide on:
[...]

PR #270, opened 2016-06-05
Pyro revamp

AfterShockFF:
This is to address a number of concerns and move the pyro towards a much more fun class to play and play against. This also has the intention of making him playable in the current competitive scene as an offensive class.

The revamp currently involves the following:
- New movement ability: Jetpack
  -- Hold right click whilst in the air to produce additional thrust (upwards and horizontally)
- Revamped burning
  -- No more afterburn damage-over-time ticks. Instead, increased burn levels will increase damage from the IC and flamethrower
  -- The flamethrower or napalm nade is now necessary to light a player on fire. After that, any fire damage (IC/flamethrower/napalm) will contribute towards increasing the burn level based on damage done.
- Removal of flamethrower push
  -- The jetpack now replaces this
  -- This also means you can use the flamethrower whilst bunny hopping, sliding etc without penalty
  -- This also allows players to fire IC projectiles at enemies whilst jetpacking
- Incendiary Cannon revamp
  -- The IC now has a clip
  -- The IC now fires and reloads much faster
  -- IC base damage increased
  -- IC jumping now does much less damage to the pyro, so can be used in conjunction with the jetpack for a variety of trick jumps
- Improved napalm grenade
  -- Slightly increased napalmlet damage
  -- Increased the vertical damage radius of napalmlets
  -- Increased the rate at which napalmlets check for players (harder to avoid damage by jumping through them)
  -- Made napalmlets distribution less random

Scripts pull request: https://github.com/fortressforever/fortressforever-scripts/pull/58

PR #298, opened 2016-08-15
Actually bring MIRV damage in line with frag gren (from 180 down to 145)

squeek502:
Was meant to be included in 2.4 but the MIRV change got missed in the merge or something, see:
- https://github.com/fortressforever/fortressforever/commit/da596605a65ff3780b837a2e61dd028fef292fd9
- https://github.com/fortressforever/fortressforever/commit/33cf7dff38bba3afba3a6a0b42d1cca08382c23c
- https://github.com/fortressforever/fortressforever/commit/6b4645ceeb4521d1c0d172180a27f29fc67b3701
- http://www.fortress-forever.com/changelogs/2.4

**Note**: The base grenade class uses damage to calculate radius (damage \* 1.5), which means this will drop the MIRV initial explosion radius from 270 to 217.5. The frag grenade has 145 damage and a manually set radius of 270. See [the first few columns of the table in this comment for how this will affect damage at the edge of the radius](https://github.com/fortressforever/fortressforever/issues/3#issuecomment-226916203)

~~Should we set the radius in line with frag grenades as well?~~ Done

### DIFF
--- a/game_shared/ff/ff_grenade_base.h
+++ b/game_shared/ff/ff_grenade_base.h
@@ -43,6 +43,9 @@
 	#define GREN_ELAS 0.4f
 	#define GREN_FRIC_CONC 0.3f
 	#define GREN_ELAS_CONC 0.7f
+
+	#define FRAG_GREN_DAMAGE 145.0f
+	#define FRAG_GREN_RADIUS 270.0f
 #endif
 
 //=============================================================================
--- a/game_shared/ff/ff_grenade_mirv.cpp
+++ b/game_shared/ff/ff_grenade_mirv.cpp
@@ -39,6 +39,7 @@
 	//static ConVar mirv_ang_z_max("ffdev_mirv_ang_z_max","0",FCVAR_FF_FFDEV,"Maximum x spawn angle for mirvlets");
 	//static ConVar mirvlet_dmg("ffdev_mirvlet_dmg","180.0",FCVAR_FF_FFDEV,"Damage a single mirvlet does");
 	#define MIRVLET_DMG 145.0f
+	#define MIRV_DMG 145.0f
 #endif
 
 class CFFGrenadeMirv : public CFFGrenadeBase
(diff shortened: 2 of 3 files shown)


Jitterskull:
Sweet... Glad to see that I actually contributed to the mod! :D

As for radius, I dunno. Either way, you wouldn't need to perfectly predict the pattern and timer of MIRV to hit right in the body of someone to do massive damage like you would have to with napalm nade just to increase afterburn level because of it tiny explosion radius, at least because it drops mirvlet. Afterall, MIRV is about taking down sg, complete area denial and panic factor grenade, using it mere as extra frag (or even more powerful version of frag) is rather silly, but I guess the nade will be ok with whether radius for it's initial explosion.

mikeparker:
Good catch guys. I think it should be the same as frag nades.

squeek502:
Radius set to the same as the frag gren.

Jitterskull:
With the reduction of the MIRV damage, does that mean demoman will receive less self damage, or it remains the same as it's now?

squeek502:
Yes, self-damage will reduce by the same percentage as base damage. You do 2/3 damage to yourself, so it'll change like so:
- Old: 180 base damage, 120 self-damage
- New: 145 base damage, 96.67 self-damage

TRUEGhostCat:
Ah, so I guess I was right that IC does more damage to self was just a bug after all. I hope that it is, at least.

squeek502:
IC self-damage being different is definitely not a bug, see [this code](https://github.com/fortressforever/fortressforever/blob/beta/game_shared/ff/ff_gamerules.cpp#L1584-L1587) (IC_SELFDAMAGEMULTIPLIER is set to 1.8)

Looks like that multiplier was put in way before my time, though, so I can't comment on why it's like that. The last time it was changed was Sep 8, 2007: https://github.com/fortressforever/fortressforever/commit/5e03aae3c6a217f3892e4a9b8856fd8e5c1bb890#diff-bdce6cbeb293db05ee565b8237ed30b1R82

However, that modifier is actually something that #270 doesn't seem to be accounting for (it introduces a _second_ self-damage modifier instead of tweaking the existing one). Will get that sorted.

AfterShockFF:
Looks good to me :+1:

PR #299, opened 2016-08-22
Fix bullets/flamethrower/etc being blocked by projectiles

### LINKED ISSUE #291: Bullet/flamethrower/etc traces are blocked by projectiles
Split from #270. @MaxRideWizardLord said:

> Currently class Pyro already suffers from many bugs and issues that hold him from it's full intended power, for example infamous and notorious hit registry issue of flamethrower, which makes pyro do not deal damage at all if the enemy is in corner, point black, under certain angle, or shoot any projectile based weapon which cancels flamethrower completely.

squeek502:
- Closes #291 

---

This is **_a**_ fix, but not necessarily a _good_ fix. The way this is written currently could have unintended side-effects, as [COLLISION_GROUP_NONE is used _a lot_ throughout the codebase](https://github.com/fortressforever/fortressforever/search?utf8=%E2%9C%93&q=COLLISION_GROUP_NONE).

An alternate solution would be to create a new COLLISION_GROUP specifically for weapon traces and make sure to use that for every relevant trace. That would be less prone to unintended side-effects (and give more fine-grained control), but we'd need to make sure that we convert everything to the new COLLISION_GROUP that should use it.

It's worth noting that bullets colliding with projectiles seems to be the default behavior of the Source SDK; as far as I can tell this behavior still exists in Source SDK 2013 (still uses COLLISION_GROUP_NONE for bullet traces, no special handling in the relevant collision functions).

### DIFF
--- a/game_shared/ff/ff_gamerules.cpp
+++ b/game_shared/ff/ff_gamerules.cpp
@@ -1962,6 +1962,17 @@ ConVar mp_friendlyfire_armorstrip( "mp_friendlyfire_armorstrip",
 
 bool CFFGameRules::ShouldCollide( int collisionGroup0, int collisionGroup1 )
 {
+	// HACK: bullet/hull traces use COLLISION_GROUP_NONE, so make them not collide
+	// with rockets/projectiles/weapons. Doing this before the sorting makes sure that
+	// normal objects can still collide
+	if (collisionGroup0 == COLLISION_GROUP_NONE && (
+		collisionGroup1 == COLLISION_GROUP_ROCKET ||
+		collisionGroup1 == COLLISION_GROUP_PROJECTILE ||
+		collisionGroup1 == COLLISION_GROUP_WEAPON))
+	{
+		return false;
+	}
+
 	// Do this before the groups are re-ordered. This way we can check only when
 	// one entity pushing on another, and not the other way round.
 	// This is checking for players moving into grenades


AfterShockFF:
> An alternate solution would be to create a new COLLISION_GROUP specifically for weapon traces and make sure to use that for every relevant trace. That would be less prone to unintended side-effects (and give more fine-grained control), but we'd need to make sure that we convert everything to the new COLLISION_GROUP that should use it.

That doesnt sound like much work, there are only a handful of weapons?
- Single shotgun
- Super shotgun
- AR
- AC
- FT
- Sniper rifle

squeek502:
> That doesnt sound like much work, there are only a handful of weapons?

True, but it seems like it should also be applied to some non-weapon traces as well (off the top of my head, things like the crosshair info trace). We'd probably have to go through every instance of traces with COLLISION_GROUP_NONE and decide if they should use the new COLLISION_GROUP instead.

Or maybe we should do it the other way around? Add a new COLLISION_GROUP and use it only for things we know we _want_ to collide with projectiles/weapons?

AfterShockFF:
Probably the SG aiming trace too. I'm OK with just testing the current fix, it shouldnt be too hard to play around and see whats changed?

squeek502:
Yeah, the current blanket fix might not actually be a problem. There was nothing obvious broken in the tiny amount of testing I did (rockets still collide with stuff, etc). I'm not really sure how common it is that we'd actually want a COLLISION_GROUP_NONE trace to collide with a rocket/projectile/weapon (seems like all the collisions we care about would go in the opposite direction, a rocket/projectile/weapon colliding with something else).

SergeantGrimes:
I know that there been said about collision with other objects, but does this fix mean that flamethrower's flame can actually pierce through enemies and hit everyone with these particles, like real flamethrower would?? Basically make it hit multiple enemies like explosives do. I'm not sure about FF entirely, but in TF2 explosive can hit enemy that even is behind other enemy, or superposed one to another as the explosive pierce through enemies.

squeek502:
@SergeantGrimes that would be separate.

squeek502:
Seems to be working alright, merging.

Issue #300, opened 2016-08-25
Jump Gun ammo issues


doublah:
Currently, jump gun has some ammo-related issues:
It uses ammo, giving it limited use.
It shows ammo/ammo type to the user (this should be removed like with melee weapons).
It shares ammo type with nailgun (currently if you use all your nailgun ammo, you cannot use the jump gun, and vice versa).

AfterShockFF:
I agree with all this - thanks for opening.

PR #301, opened 2016-08-27
Remove ammo usage in jumpgun code

squeek502:
Not sure if it's necessary for https://github.com/fortressforever/fortressforever-scripts/pull/61 but seems worthwhile regardless.

Contributes towards #300


AfterShockFF:
Not sure what I think about this. I guess if we want to reinstate ammo usage at some point this code could come in handy. That said we can just come back to this pull request and revert it..! :+1:

squeek502:
Turns out this is probably necessary (as far as I can tell, with primaryAmmoType == -1, GetAmmoCount returns 0 so it would always act as if it was out of ammo).

PR #302, opened 2016-08-27
hud_weaponinfo safety and improvements

squeek502:
- Fixed crash when a weapon's ammo or weapon icon is null
- Reduced the number of copy constructors used (only copy in the icons when the weapon is different than last tick)
- Fixed memory leak (2 new CHudTexture's on every map load that would never get deleted)

Necessary for https://github.com/fortressforever/fortressforever-scripts/pull/61


AfterShockFF:
Seems cool :+1:

PR #314, opened 2016-09-11
Fix #313 pyro IC bonus damage kills with 'player'

### LINKED ISSUE #313: Pyro IC bonus damage sometimes kills with 'player'
This is because the bonus damage isn't flagged as IC damage correctly.

AfterShockFF:


squeek502:
Wow, surprised it worked at all before; the inflictor was set to CFFGameRules!

:+1:

PR #317, opened 2016-09-11
Fix voice icons on scoreboard

### LINKED ISSUE #312: Voice Assert when 22 players on server
Some of the voice code is off-by-one in the voice array, it uses VOICE_MAX_PLAYERS which is 22 but it starts at 1. 

Doesn't cause crashes but it probably does something weird like not displaying when the 22nd person on the server talks.

Code here, you can see mirv actually fixed this once already. I dont know if mulch accidentally broke it again

https://github.com/fortressforever/fortressforever/blob/beta/cl_dll/game_controls/ClientScoreBoardDialog.cpp#L770-L775

If this code is wrong, everyones voice icon will be off by 1 row. Should be easy to check i guess!

AfterShockFF:
displaying the status of voice icons from someone else, and asserting when the 22nd player talks.

Closes #312

### DIFF
--- a/game_shared/voice_status.cpp
+++ b/game_shared/voice_status.cpp
@@ -326,9 +326,9 @@ int CVoiceStatus::GetSpeakerStatus(int entindex)
 	if ( !engine->GetPlayerInfo( entindex, &pi ) )
 		return 0;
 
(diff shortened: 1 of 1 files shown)


squeek502:
:+1:

PR #318, opened 2016-09-11
Fix assists not being logged to file properly

squeek502:
- Was treating the player entindex as a user id


squeek502:
Closing in favor of #319 which will also fix this.

PR #319, opened 2016-09-11
Assists: send user ID over the network rather than entindex

squeek502:
Alternate solution for #318

### DIFF
--- a/cl_dll/ff/ff_hud_deathnotice.cpp
+++ b/cl_dll/ff/ff_hud_deathnotice.cpp
@@ -706,14 +706,12 @@ void CHudDeathNotice::FireGameEvent( IGameEvent * event )
 		deathMsg.iSuicide = ( !killer || ( ( killer == victim ) && ( !bBuildableKilled ) ) );
 
 		deathMsg.Assister.iEntIndex = -1;
-		int killassisterIdx = event->GetInt( "killassister", -1 );
-		if (killassisterIdx != -1)
+		int killAssister = event->GetInt( "killassister", -1 );
+		if (killAssister != -1)
 		{
-			//C_BaseEntity* pEnt = cl_entitylist->GetEnt( killassisterIdx );
-			//C_FFPlayer *pFFPlayer = ToFFPlayer( pEnt );
-			//if ( !pFFPlayer )
-			Q_strcpy( deathMsg.Assister.szName, g_PR->GetPlayerName( killassisterIdx ) );
-			deathMsg.Assister.iEntIndex = killassisterIdx;
+			killAssister = engine->GetPlayerForUserID(killAssister);
+			Q_strncpy( deathMsg.Assister.szName, g_PR->GetPlayerName( killAssister ), MAX_PLAYER_NAME_LENGTH );
+			deathMsg.Assister.iEntIndex = killAssister;
 		}
 
 		// only consider suicide if there was no kill assist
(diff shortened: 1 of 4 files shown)


cl_dll/ff/ff_hud_deathnotice.cpp
```diff
@@ -706,14 +706,12 @@ void CHudDeathNotice::FireGameEvent( IGameEvent * event )
...
-		if (killassisterIdx != -1)
+		int killAssister = event->GetInt( "killassister", -1 );
+		if (killAssister != -1)
 		{
-			//C_BaseEntity* pEnt = cl_entitylist->GetEnt( killassisterIdx );
-			//C_FFPlayer *pFFPlayer = ToFFPlayer( pEnt );
-			//if ( !pFFPlayer )
-			Q_strcpy( deathMsg.Assister.szName, g_PR->GetPlayerName( killassisterIdx ) );
-			deathMsg.Assister.iEntIndex = killassisterIdx;
+			killAssister = engine->GetPlayerForUserID(killAssister);
```
AfterShockFF: I think this would be more explicit if we keep the concepts of userId and playerIndex separate and stored in different variables. It would help prevent future coders getting confused and making a bug.

killAssisterUserId and killAssisterIndex for example.

game_shared/multiplay_gamerules.cpp
```diff
@@ -810,7 +810,7 @@ bool CMultiplayRules::IsMultiplayer( void )
 				RecentAttackerInfo *pTopAssister = pFFPlayer->GetTopKillAssister( pScorer );
 				if ( pTopAssister )
 				{
-					event->SetInt("killassister", pTopAssister->playerIndex );
+					event->SetInt("killassister", pTopAssister->GetUserID() );
```
AfterShockFF: Does this compile? I'm confused - RecentAttackInfo doesnt have a GetUserId function on?

squeek502: Haha I'm a dummy

dlls/ff/ff_player.cpp
```diff
@@ -7908,14 +7908,13 @@ void CFFPlayer::AddRecentAttacker( const CTakeDamageInfo &dmgInfo )
...
 
-	int attackerIdx = pAttacker->entindex( );
 	float dmg = dmgInfo.GetDamage( );
 	float timestamp = gpGlobals->curtime;
 
 	// search for existing or create new
 	for ( int i = 0; i < m_recentAttackers.Count( ); ++i )
 	{
-		if ( m_recentAttackers[i].playerIndex == attackerIdx || pAttacker == m_recentAttackers[i].pFFPlayer )
+		if ( pAttacker == m_recentAttackers[i].hPlayer )
```
AfterShockFF, line 7917: You don't have to do a .Get here then?

dlls/ff/ff_player.h
```diff
@@ -102,18 +102,15 @@ struct LocationInfo
...
 	float totalDamage; // note: before armor scale. dont really care. just tracks raw damage received
 	float timestamp;
 
-	CFFPlayer *pFFPlayer;
-
-	RecentAttackerInfo( int idx, float dmg, float ts, CFFPlayer *pPlayer)
+	RecentAttackerInfo(CFFPlayer *pPlayer, float dmg, float ts)
 	{
-		playerIndex = idx;
+		hPlayer = pPlayer;
```
AfterShockFF, line 111: Does this casting happen by magic then?

dlls/ff/ff_player.cpp
squeek502: CHandle has an == override that makes this work

dlls/ff/ff_player.h
squeek502: CHandle has an assignment override that makes this work

Issue #309, opened 2016-09-11
Pyro jetpack sound plays multiple times sporadically


squeek502:

squeek502:
Did a bit of testing to see what's causing this. It seems like what's going on is for each currently jetpacking pyro in the server, the sound gets doubled. If anyone is certain that they have gotten this bug when they were the only pyro in the server, let me know.

Why exactly it gets louder on your client I can't really figure out, though, as it _should_ just be playing multiple sounds at the location of each of the jetpacking players.

It could be down to the soundscript definition, maybe the sound file, or it could be how it's being emitted in code. As far as I can tell, the sound is being emitted in a similar method to the flamethrower sound (although that's being emitted in shared code, while the jetpack sound is emitted in client code).

PR #329, opened 2016-09-19
Fix jetpack effect staying on when you switch class while jetpacking

squeek502:
- If this is released, it will cause 'Server uses different class tables' until the servers are updated

The way this worked was likely a misguided attempt at fixing #309 but without a real understanding of what was actually happening. Syncing bJetpacking to the client shouldn't be able to cause doubling of anything anymore (if it ever did).


DexterHaslem:
a nice trivial fix.. blast away... see what i did there half a year later

Issue #311, opened 2016-09-11
Crashing after latest patch (v2.7)


AfterShockFF:
Need to download some Steam/*.dmp files from players to get to the bottom of this.

squeek502:
Closing under the assumption that the client crashes introduced by 2.7 have been fixed. Correct me if I'm wrong.

PR #330, opened 2016-09-19
Add null pointer safety to fx_bonusfire

squeek502:
The entity sent over the network in the effect data could potentially not exist once it reaches the client, which would cause a null pointer crash

Contributes to #311

### DIFF
--- a/cl_dll/ff/ff_fx_bonusfire.cpp
+++ b/cl_dll/ff/ff_fx_bonusfire.cpp
@@ -50,7 +50,11 @@ void BonusFireCallback( const CEffectData &data )
 
 	Vector	projectileOrigin = data.m_vOrigin;
 	float	scale = data.m_flScale;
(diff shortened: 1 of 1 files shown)


squeek502:
Confirmed that this can occur:
- Added this code: `engine->ServerCommand(UTIL_VarArgs("kick %s\n", this->GetPlayerName()));` right after [DispatchEvent("BonusFire", data) in ff_player.cpp](https://github.com/fortressforever/fortressforever/blob/beta/dlls/ff/ff_player.cpp#L5287)
- Set net_fakelag to 200 (not sure if this is needed)
- Added a bot, set him on fire and shot him with an IC to trigger the bonus fire effect
- Bot gets kicked, ent index is received on the client and assumed to exist -> null pointer crash

Going to merge and release another hotfix since we _know_ this is another crash.

PR #341, opened 2018-01-23
Add discord rich presence

DexterHaslem:
Just for fun, show some info in discord rich presence. This will require the 'discord-rpc.dll' to be added to builds. however if it is not present, it will not cause issues.

Currently for details it shows the following in the discord player card:

```
<team> <team points> - <classname> <points> <k>:<d>
[1/16] server hostname blah blah
```

wonk things: for discord to properly initialize it needs to run frames when not in game, so i abused 
`CHLClient::HudUpdate()` which runs even in the main menu. alternatively, since we cant really use the join functionality, we can run it only in client player think, too (how i originally had it)

### DIFF
--- a/cl_dll/cdll_client_int.cpp
+++ b/cl_dll/cdll_client_int.cpp
@@ -86,6 +86,7 @@
 #include "avi/iavi.h"
 #include "hltvcamera.h"
 #include "ff_vieweffects.h"
+#include "ff_discordman.h"
 
 // BEG: Added by Mulchman for team menu at level start up
 #include <cl_dll/iviewport.h>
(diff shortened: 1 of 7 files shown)


doublah:
Couldn't the client send the server connection info to a joining client?

Also does it show the map?

DexterHaslem:
Yes, but the way discord works, whoever you joined on would get a popup asking if you want them 'to join you' because its designed for party lobbies. It could be hardcoded to auto accept, but its still wonky and kinda pointless.
Even further, this feature requires approval by discord and hashing of the join info which doesn't really make sense to join a server ip

ill put map name in server info line. was going to do thumbnails but that would require upkeep

PR #347, opened 2018-01-26
Fix laggy jetpack fuel

### LINKED ISSUE #325: Jetpack fuel bar is glitchy on high ping
Could do with some client-side prediction or something.

DexterHaslem:
Changed jetpack fuel from float to int to make it easier to work with (and pdump),
added client prediction to fuel, recharge rate (and class next ability so its responsive on high pings)

fixes #325


dlls/ff/ff_player.cpp
```diff
@@ -8024,12 +8024,12 @@ void CFFPlayer::RemoveMeFromKillAssists( )
...
 {
-	return m_flJetpackFuel;
+	return m_iJetpackFuel;
 }
 
-void CFFPlayer::SetJetpackFuel(float newFuel)
+void CFFPlayer::SetJetpackFuel(int newFuel)
 {
-	m_flJetpackFuel = max(100.0f, newFuel);
+	m_iJetpackFuel = max(200, newFuel);
```
squeek502: Note: This value will need to get updated in the base_teamplay Lua as well (https://github.com/fortressforever/fortressforever-scripts/pull/64).

The alternative would be to change this function to SetJetpackFuelPercent (either by name or just in terms of functionality), so that `SetJetpackFuel(100)` always sets fuel to 100% regardless of the internal representation. Doing that might be better?

game_shared/ff/ff_player_shared.cpp
```diff
@@ -757,11 +761,19 @@ void CFFPlayer::ClassSpecificSkillHold()
 	switch (GetClassSlot())
 	{
 		case CLASS_PYRO:
+//			note this is not needed, we *do* want to reduce fuel on prediction
```
squeek502: Can remove these comments and the added `#include "prediction.h"` if they're not needed.

DexterHaslem:
made things maybe less stupid

squeek502 (approved):

Jitterskull:
Can I sincerely ask you all though, did jetpack get stealth tweaks?? Like, was the fuel total amount\fuel consuption and fuel recharge (regen) changed? It feels like it regenerates the fuel much slower now and playing as pyro feels overall very uncomfortable, unfamiliarly and frustrating. Often when I feel like I could make certain lenght jumps using memorized timing on how much I need to wait certain amount of time to be sure that I could make the jump based on my previous experience, result in failure and that really bothers me. After all that, I still get jittery jetpack meter in hud on a pingy servers.

Haven't played in a while, so it might be my perceptions deceiving me, but it sure feels really annoying and too slow to regen the fuel back to do just a single high jump using jetpack. Would like it at least be the way it was.

DexterHaslem:
Not intentionally, fuel regen was intended to be same, just switched from float to int to be easier for network prediction and more efficient for netcode. I double checked - 
old:
1.0 fuel every frame to max of 100 (100 steps)

new: 
1 fuel every frame to max of 200 (200 steps)

Accidentally half'd the jetpack refuel rate by forgetting to double increment rate 😢 

https://github.com/fortressforever/fortressforever/commit/93026c6b1a6db27f5fd0f5d8c1ea756549eb99aa#diff-7234bbde41b43c96e5a259fb6924b463L1830

fixed in #354

Jitterskull:
Glad that my comment was useful! :D

So what about the fuel consuption, it feels EXACTLY the same as it was before this update?

PR #362, opened 2018-10-06
add more debug instructions to README

DexterHaslem:
quickie , I noticed the readme was missing instructions on how to debug directly from VC++, so added a section on that, and removed some outdated dll copy instructions while in there.


squeek502 (approved):
Nice, I've always just used Attach to process after launching. :no_mouth:

alaswell (approved):

Issue #273, opened 2016-06-05
Maliciously sabotaged SG explosion damage can kill teammates when friendlyfire is off


squeek502:
If a maliciously sabotaged sentry gun explodes (not sure if it was killed or if the sab ended), then the explosion damage can hurt the spy's teammates even when friendlyfire is off. Saw this today, thought it might be a bug.

squeek502:
Seems like this is probably caused because the owner is always set as the attacker and the SG as inflictor in the CTakeDamageInfo in CFFSentryGun::DoExplosionDamage. Should probably set the attacker to the saboteur if the SG is maliciously sabotaged.

Relevant code:
- https://github.com/fortressforever/fortressforever/blob/beta/dlls/ff/ff_sentrygun.cpp#L1915

PR #363, opened 2018-10-06
Makes sabotaged things respect friendly fire setting

### LINKED ISSUE #327: Sabotaging spy (and their team) can damage own sabotaged SG if friendly fire is off

DexterHaslem:
This makes sentry gun and dispenser damage given/received check friendly fire is on in respect to the saboteur's team. fixes #327 

I cannot spell saboteur correctly

### DIFF
--- a/game_shared/ff/ff_gamerules.cpp
+++ b/game_shared/ff/ff_gamerules.cpp
@@ -2096,77 +2096,41 @@ bool CFFGameRules::FCanTakeDamage( CBaseEntity *pVictim, CBaseEntity *pAttacker 
 	// this fixes #327 , which was SG specific, but applied same
 	// treatment to detting dispensers
 
-	// If an SG is shooting its teammates then allow it to hurt them
-	if (pAttacker && pAttacker->Classify() == CLASS_SENTRYGUN)
-	{
-		CFFSentryGun *pSentry = dynamic_cast< CFFSentryGun* > (pAttacker);
-
-		if ( pSentry && pSentry->IsMaliciouslySabotaged() )
-		{
-			bool victimIsSabTeammate = FFGameRules()->IsTeam1AlliedToTeam2( 
-				pVictim->GetTeamNumber(), 
-				pSentry->m_iSaboteurTeamNumber ) == GR_TEAMMATE;
-
-			return victimIsSabTeammate ? isFriendlyFireOn : true;
-		}
-	}
+	// if its a buildable, let it damage/be damaged by players based on team & friendly fire 
 	
-	if ( pVictim && pVictim->Classify() == CLASS_SENTRYGUN )
+	CFFBuildableObject *pBuildableAttacker = dynamic_cast <CFFBuildableObject *> (pAttacker);
+
+	if ( pBuildableAttacker && pBuildableAttacker->IsMaliciouslySabotaged() )
 	{
-		CFFSentryGun *pSentry = dynamic_cast <CFFSentryGun *> (pVictim);
-
-		// Allow team to kill their own SG if it is sabotaged
-		if ( pSentry && pSentry->IsSabotaged() )
-		{
-			bool attackerIsSabTeammate = FFGameRules()->IsTeam1AlliedToTeam2( 
-				pAttacker->GetTeamNumber(), 
-				pSentry->m_iSaboteurTeamNumber ) == GR_TEAMMATE;
-
-			return attackerIsSabTeammate ? isFriendlyFireOn : true;
-		}
-
- 		// if it's not sabotaged then we need to get its owner and use it later on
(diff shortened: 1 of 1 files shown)


DexterHaslem:
this might cover #273 as well, I will check if that explosion code checks the same gamerules code

DexterHaslem:
oh great, this doesnt work for malicious sabotage of SG because no saboteur info is set in that case to distinguish them. I need to change that code to have a proper flag and still keep track of sab team first.. this is WIP

squeek502:
I'm not really seeing how #327 was happening in the first place. It looks like the previous code was essentially, "if an SG is sabotaged, anyone can damage it," and [`IsSabotaged()` is a prereq of `IsMaliciouslySabotaged()`](https://github.com/fortressforever/fortressforever/blob/a4ba59abe2cfdef801c893155a95f0debbcf4c4e/dlls/ff/ff_buildableobject.cpp#L602-L617), which means that `IsSabotaged()` must include maliciously sabotaged SGs, right? So I'm confused. Did you confirm #327 was real?

game_shared/ff/ff_gamerules.cpp
```diff
@@ -2087,27 +2087,45 @@ bool CFFGameRules::FCanTakeDamage( CBaseEntity *pVictim, CBaseEntity *pAttacker
 	// if it's a buildable, then we use the buildable's owner to perform the team checks etc. -> Defrag
 	CBasePlayer *pBuildableOwner = NULL;
 
+	bool isFriendlyFireOn = friendlyfire.GetInt() != 0;
```
squeek502: `GetBool()` would make more sense here

game_shared/ff/ff_gamerules.cpp
```diff
@@ -2087,27 +2087,45 @@ bool CFFGameRules::FCanTakeDamage( CBaseEntity *pVictim, CBaseEntity *pAttacker
...
 	}
 	
 	if ( pVictim && pVictim->Classify() == CLASS_SENTRYGUN )
 	{
 		CFFSentryGun *pSentry = dynamic_cast <CFFSentryGun *> (pVictim);
 
 		// Allow team to kill their own SG if it is sabotaged
-		if (pSentry && pSentry->IsSabotaged())
-			return true;
+		if ( pSentry && pSentry->IsSabotaged() )
```
squeek502: Haven't thought this through fully, but it seems like it should be possible to condense most/all of these if blocks into a single block and treat all buildables the same as appropriate, right? Check if `Classify() == CLASS_SENTRYGUN || Classify() == CLASS_DISPENSER`, then just cast to `CFFBuildableObject` or w/e, and all the code inside the if could be the same.

squeek502 (changes requested):

game_shared/ff/ff_gamerules.cpp
DexterHaslem: yes, strongly agree. I would have made it shorter but i didnt spend the time. Ill try to

game_shared/ff/ff_gamerules.cpp
DexterHaslem: doh!

DexterHaslem:
It definitely exists (just sab a SG then shoot it), thats what i was trying to address. I assumed the sabotage properties would always be set on the buildable base class, which is not true, I think. I will re-test

squeek502:
Where does FCanTakeDamage return false in that instance (when trying to damage your maliciously sabotaged SG)? Or is it something else that's affecting things?

DexterHaslem:
The previous code always returned true if it was sabotaged, is that what you are asking?

squeek502:
Yeah, but if that's true, shouldn't it be impossible for there to be a bug where a (maliciously) sabotaged SG couldn't be damaged? Maybe I'm misunderstanding what #327 is saying?

DexterHaslem:
Indeed, it's saying it could be damaged by the sabotuer which arguably doesn't make sense when friendly fire is on. this really comes down to a gameplay decision:

do sabotaged buildings 'belong' to the sabotaging team? I would not be surprised if the original devs intentionally made them 'building versus the world' when sabd

squeek502:
Ohhh, ok I get it. The spy (and their team) can damage the gun they have maliciously sabotaged when FF is off. The title of #327 ~~is~~ (edit: was) super confusing.

DexterHaslem:
True that! I'm traveling and will finish this up after next week

DexterHaslem:
I debugged through this, didnt understand how sabotaging works - `m_hSaboteur` is always set so disregard that. I will refactor the duplicated sentry/dispenser stuff then this should be good to go

DexterHaslem:
just tested it , works more gooder. also learned you can safely pass NULL to dynamic cast, so i'm just doing that to check it's a buildable

squeek502 (approved):

PR #366, opened 2019-01-27
Replace some explicit dynamic_cast's with `ToX` utility functions

squeek502:
Slight optimization. The `ToX` utility functions use virtual base class functions along with static_cast in non-debug builds


hanoicity900:
lam sao cho trang cua minh len duoc nhi
http://www.dackhuvietnam.com/2019/01/ban-biet-thu-lien-ke-van-canh-an-lac.html

PR #370, opened 2019-07-01
Fix first gren after death sometimes being thrown immediately

squeek502:
m_bWantToThrowGrenade wasn't being reset on death/spawn, so if you die while throwing grens it would sometimes get stuck on and the first gren primed after spawning would immediately get thrown

This was a long-standing and very annoying bug


AdamWillden:
@squeek502 amazing this wasn't found sooner. like 10 years ago 😆

squeek502:
Every time I look at FF code I'm amazed at how much of a mess it is. These types of mistakes are everywhere.

PR #373, opened 2019-09-01
Changes the way the spies name is displayed when disguised

### LINKED ISSUE #368: Spy disguise name choice
From Discord:

> Have you considered adding the name for the person when disguised, so they know who to avoid.

me: 

> havent considered that about spy, would make sense

> hm, actually looking at the code the name shown could possibly be different between players
so that would need to change too

> it kind of does things in a silly way, could definitely be improved
if there's a player that is the class you're disguised as on the team, then the name shown is always the first match found, if there's no player with the class then it chooses a random player on that team
not very spy-friendly since it doesn't randomly pick between multiple players with the same class

Relevant code:
https://github.com/fortressforever/fortressforever/blob/e7b488a008cb79b92be4a4efdea42899e8f7d1e8/cl_dll/ff/ff_hud_crosshairinfo.cpp#L390-L442

alaswell:
This is completely untested, but should resolve #368

### DIFF
--- a/cl_dll/ff/ff_hud_crosshairinfo.cpp
+++ b/cl_dll/ff/ff_hud_crosshairinfo.cpp
@@ -391,11 +391,12 @@ void CHudCrosshairInfo::OnTick( void )
 									// that is playing the class this guy is disguised as
 
 									// Gonna generate an array of people on the team we're disguised as
-									// in case we have to randomly pick a name later
-									int iPlayers[ 128 ], iCount = 0;
-
+									// and a sub-array of people who share the same class we're disguised as
+									// so that we can randomly pick a name later
+									int iPlayers[ 128 ], iPlayersWithSameClass[ 128 ], iCount = 0, iSameClassCount = 0;
+	
 									bool bDone = false;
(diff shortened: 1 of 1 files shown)


squeek502 (approved):
Looks good.

Note: this is only part of #368--the other part would be syncing the chosen player across all clients so that each person sees the same name, which would also allow us to show the spy which person they are disguised as.

Right now the name is/will be different between each client, but syncing it would require changes to the networking tables, which would cause a 'server uses different class tables' error and force servers to update which isn't ideal since servers can be pretty slow to update these days.

alaswell:
Yes, good point on the other part of #368 

We will have to add that in if/when the networking tables change again for something more significant.

doublah:
How does this look from spectator? Is it changed there?

squeek502:
Do you know if crosshairinfo shows while spectacting? I would think this stops it from showing:

https://github.com/fortressforever/fortressforever/blob/28b1c139647f2a68baa03f4eadd4a0eed708ed0f/cl_dll/ff/ff_hud_crosshairinfo.cpp#L143-L144

PR #376, opened 2020-05-01
Nerfing the frag grenade (needs more testing)

pvorisek25:
After playing several matches and reading several posts from the community I found out that the frag grenade is OP (wow). I nerfed the frag grenade's damage from 145 to 100 (69% of the original damage) and the frag grenade's radius from 270 to 175 (65% of the original radius). However more testing needs to be done on this topic.

### DIFF
--- a/game_shared/ff/ff_gamerules.cpp
+++ b/game_shared/ff/ff_gamerules.cpp
@@ -1229,8 +1229,14 @@ ConVar mp_friendlyfire_armorstrip( "mp_friendlyfire_armorstrip",
 		//NDebugOverlay::Cross3D(vecSrc, 8.0f, 255, 0, 0, true, 5.0f);
 #endif
 
+		falloff = info.GetFallOff();
+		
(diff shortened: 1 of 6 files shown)


squeek502:
Changing the frag grenade damage is probably not going to happen. If I remember right, it matches TFC, and the radius is lower than TFC. However, if we were to touch the frag grenade radius, I'd mostly want to address https://github.com/fortressforever/fortressforever/issues/3.

pvorisek25:
I've looked at the #3 issue and at the actual code and I think that @mikeparker's option 4 in the #3 issue would be the best way to go. 
> Adjustable falloff, adjustable radius (the edge will be whatever % we want, and the fall rate can be as fast or slow as we want)

Since we would have more grenade properties to play with, we would have more ways of balancing grenades and more space for new ones. What do you think about that?

DexterHaslem:
is the grenade cone visualization thinger still in? i cant even remember, but it is very useful for this

PR #379, opened 2020-10-29
Remove ternary expressions

AdamWillden:
I came across these and wanted to tidy them up.

straight `? true : false` I removed using find replace.
inverted `? false : true` I went through manually to resolve.

I'm okay if you don't want to do the inverted ternary expressions but I don't see why not. Check the individual commit itself to review these.


AdamWillden:
Ugh, I see bogus 'performance warnings' I'll review this.

squeek502:
Perhaps a slightly strange reason, but one possible reason for not doing this is that it would reduce the legibility/relevance when diffing against the base SDK code ([example from when we were attempting to port to the 2013 SDK](http://www.fortress-forever.com/diff/)).

It is weird that this is a common pattern in the SDK code, though; I never realized that.

AdamWillden:
Mmmm as dumb as it is to me and as much as I’d rather a ‘!!(integer/pointer)’ notation it’s not worth changing.

PR #380, opened 2020-10-31
Remove ternary expressions in FF code

AdamWillden:
I've decided to drop the non-FF stuff and keep the FF changes, but it's up to you @squeek502.

A few of the FF ones were entirely pointless.


squeek502 (approved):
Sounds good to me

PR #381, opened 2020-10-31
Misc changes

AdamWillden:
Some changes in preparation of the HUD update - to keep existing unrelated changes separate.

### DIFF
--- a/cl_dll/c_playerresource.h
+++ b/cl_dll/c_playerresource.h
@@ -11,7 +11,6 @@
 #pragma once
 #endif
 
-#include "shareddefs.h"
 #include "const.h"
 #include "c_baseentity.h"
 #include <igameresources.h>
--- a/dlls/ff/ff_player.cpp
+++ b/dlls/ff/ff_player.cpp
@@ -494,10 +494,6 @@ ConCommand cc_CreatePredictionError( "CreatePredictionError", cc_CreatePredictio
 // --------------------------------------------------------------------------------
 CFFPlayer::CFFPlayer()
 {
-#ifdef FF_BETA_TEST_COMPILE
-	CBaseEntity *p = NULL;
-	p->Activate();
-#else
 	m_PlayerAnimState = CreatePlayerAnimState( this, this, LEGANIM_9WAY, true );
 
 	UseClientSideAnimation();
@@ -613,8 +609,6 @@ CFFPlayer::CFFPlayer()
 
 	m_recentAttackers.Purge();
 	//m_iStatsID = -1;
-
-#endif // FF_BETA_TEST_COMPILE
 }
 
 CFFPlayer::~CFFPlayer()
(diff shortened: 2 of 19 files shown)


AdamWillden:
These were just minor changes I did along the way. I may have a secondary PR to this but need investigation first.

cl_dll/c_playerresource.h
```diff
@@ -1,4 +1,4 @@
-//========= Copyright � 1996-2005, Valve Corporation, All rights reserved. ============//
```
squeek502: ?

dlls/ff/ff_player.cpp
```diff
@@ -1708,15 +1666,10 @@ void CFFPlayer::SetupClassVariables()
 	}
 
 	ClearSpeedEffects();
-
-	// clear recent attacker (kill assist) info
-	m_recentAttackers.Purge();
```
squeek502, line 1713: This looks relevant on first glance--why is it removed here?

game_shared/ff/ff_utils.cpp
```diff
@@ -250,76 +250,57 @@ const char *FF_GetDefaultWeapon( const char *classname )
...
+			innerCol = ColorFade( iAmount, iYellow, iGreen, INTENSITYSCALE_COLOR_YELLOW, INTENSITYSCALE_COLOR_GREEN );
+		else
+			innerCol = INTENSITYSCALE_COLOR_GREEN;
 	}
+	else
+	{
+		innerCol = INTENSITYSCALE_COLOR_DEFAULT;
+	}
+
+	return *new Color(innerCol.r(), innerCol.g(), innerCol.b(), iAlpha);
```
squeek502: Not specific to this PR since it was doing this previously as well, but won't this leak the Color? Why isn't this just returning a stack-allocated Color?

AdamWillden: I will address none the less if I figure out what it is that needs to change. Didn’t look too closely at the time. I’ll ask if I’m unsure

dlls/ff/ff_player.cpp
AdamWillden: Rebase conflict accident I suspect

cl_dll/c_playerresource.h
AdamWillden: I will restore this and other instances I see

game_shared/ff/ff_utils.cpp
squeek502: Should be as simple as

```suggestion
	return Color(innerCol.r(), innerCol.g(), innerCol.b(), iAlpha);
```

AdamWillden: Cool, I’ve not done C++ since my last development 10years ago. So the oddities of it are lost on me after years of C# and JavaScript

dlls/ff/ff_player.cpp
AdamWillden: Not sure why this isn't showing as outdated...

game_shared/ff/ff_utils.cpp
AdamWillden: @squeek502 fixed in commit https://github.com/fortressforever/fortressforever/pull/381/commits/010e5a5af60ccef3c071d4f696893c8a2723cae3

AdamWillden:
@squeek502 all issues fixed in their relevant commits except the new commit https://github.com/fortressforever/fortressforever/pull/381/commits/010e5a5af60ccef3c071d4f696893c8a2723cae3 

Nothing else added or changed.

squeek502 (approved):
Looks good. :+1:

PR #388, opened 2022-12-11
Buildable changes and new Detpack Lua methods

YoYo178:
- Changed how the lua function `trigger:onbuild()` behaved
    - Individual buildables now controllable by lua.
    - `trigger:onbuild()` no longer returns a player object, instead returns what's being built.
    - Now Detpacks can be controlled even by their fuse time, (example - detpack for fuse time 10 is accepted, but any other is declined)

- Added detpack methods for lua
    - `detpack:GetFuseTime()` - Returns the fuse time of the detpack.
    - `detpack:GetDetonateTime()` Returns the exact game time the detpack will detonate on.
    - `detpack:LastFiveSeconds()` Retruns a boolean value, will return `true` if only five seconds are remaining for the detpack to explode.
    
so, i request the devs to check the code at least 10000 times because i don't know C++ at all and this is my first time contributing to an open source repository, so i might have done something that isn't visible but is in the background or something, i guess.
peace ✌️

(edit - the `PostBuild.bat` update on the first commit was unintentional, so had to make another commit to revert it.)

### DIFF
--- a/dlls/ff/ff_detpack.cpp
+++ b/dlls/ff/ff_detpack.cpp
@@ -293,8 +293,7 @@ CFFDetpack *CFFDetpack::Create( const Vector &vecOrigin, const QAngle &vecAngles
 	pObject->VPhysicsInitNormal( SOLID_VPHYSICS, pObject->GetSolidFlags(), true );
 
 	// Spawn the object
-	//pObject->Spawn();
-	// commented out for better nobuild functions, will be spawned ONLY IF it's allowed to, from ff_player.cpp
+	pObject->Spawn( );
 
 	return pObject;
 }
--- a/dlls/ff/ff_detpack.cpp
+++ b/dlls/ff/ff_detpack.cpp
@@ -293,7 +293,8 @@ CFFDetpack *CFFDetpack::Create( const Vector &vecOrigin, const QAngle &vecAngles
 	pObject->VPhysicsInitNormal( SOLID_VPHYSICS, pObject->GetSolidFlags(), true );
 
 	// Spawn the object
-	pObject->Spawn( );
+	//pObject->Spawn();
+	// commented out for better nobuild functions, will be spawned ONLY IF it's allowed to, from ff_player.cpp
 
 	return pObject;
 }
--- a/dlls/ff/ff_dispenser.cpp
+++ b/dlls/ff/ff_dispenser.cpp
@@ -414,8 +414,7 @@ CFFDispenser *CFFDispenser::Create( const Vector &vecOrigin, const QAngle &vecAn
 	//pObject->VPhysicsInitNormal( SOLID_VPHYSICS, pObject->GetSolidFlags(), true );
 
 	// Spawn the object
-	//pObject->Spawn();
-	// commented out for better nobuild functions, will be spawned ONLY IF it's allowed to, from ff_player.cpp
+	pObject->Spawn();
 
 	return pObject;
 }
--- a/dlls/ff/ff_dispenser.cpp
+++ b/dlls/ff/ff_dispenser.cpp
@@ -414,7 +414,8 @@ CFFDispenser *CFFDispenser::Create( const Vector &vecOrigin, const QAngle &vecAn
 	//pObject->VPhysicsInitNormal( SOLID_VPHYSICS, pObject->GetSolidFlags(), true );
 
 	// Spawn the object
-	pObject->Spawn();
+	//pObject->Spawn();
+	// commented out for better nobuild functions, will be spawned ONLY IF it's allowed to, from ff_player.cpp
 
 	return pObject;
 }
--- a/dlls/ff/ff_entity_system.h
+++ b/dlls/ff/ff_entity_system.h
@@ -17,6 +17,8 @@
 #pragma once
 #endif
 
+class CFFLuaSC;
+
 /////////////////////////////////////////////////////////////////////////////
 // CFFEntitySystemHelper
 /////////////////////////////////////////////////////////////////////////////
@@ -52,6 +54,7 @@ private:
 bool FFScriptRunPredicates( CBaseEntity *pEntity, const char *pszFunction, bool bExpectedVal );
 bool FFScriptRunPredicates( CBaseEntity *pEntity, const char *pszFunction, bool bExpectedVal, Vector vecOrigin, Vector vecMins = Vector(-16,-16,-16), Vector vecMaxs = Vector(16,16,16) ); // UTIL_EntitiesInBox
 bool FFScriptRunPredicates( CBaseEntity *pEntity, const char *pszFunction, bool bExpectedVal, Vector vecOrigin, float flRadius = 16 ); // UTIL_EntitiesInSphere
+bool FFScriptRunPredicates( CFFLuaSC *pContext, const char *pszFunction, bool bExpectedVal, Vector vecOrigin, float flRadius = 16 ); // UTIL_EntitiesInSphere
 
 /////////////////////////////////////////////////////////////////////////////
 #endif // FF_ENTITY_SYSTEM_H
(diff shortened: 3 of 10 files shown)


squeek502:
I think this can be done in a more backwards compatible way:

Instead of calling `onbuild` and passing in the buildable entity, `onbuild` could instead be called with an additional a parameter that contains info about the buildable (`m_iWantBuild`, `m_iDetpackTime`, etc; this could either be a plain Lua table with keys for each value (could use `luabind::adl::object` I think) or a new class that's exposed to Lua that contains this info)
  + The current `if( IsInNoBuild() && ... )` conditional would just become `if( IsInNoBuild() )`
  + The values used by `m_iWantBuild` (`FF_BUILD_DISPENSER`, etc) would probably need to be exposed to Lua (see [here](https://github.com/fortressforever/fortressforever/blob/4d8fb70b0ebcaddff59f7d3de0dc3a890ad4909c/dlls/ff/ff_lualib_constants.cpp#L266-L275) for an example)
  + The `IsInNoBuild` function would need to use a custom version of `FFScriptRunPredicates` that can provide more than just the `pEntity` in the function context, maybe a version that takes a `CFFLuaSC*` instead of a `CBaseEntity*` that the current `FFScriptRunPredicates` version could then call to avoid some code duplication.
  + the `CFFLuaSC` that `IsInNoBuild` would want to use would look something like `CFFLuaSC hOutput( 2, (CBaseEntity*)this, eitherLuabindAdlObjectOrSomeLuaBuildableInfoClass );`

With this implementation, from the Lua side it could look something like this to disallow detpacks with a certain fuse time:

```lua
function your_trigger:onbuild(player, buildable_info)
  if buildable_info.type == BuildableTypes.kDetpack and buildable_info.fusetime < 30 then
    return false
  end
  return true
end
```

and all existing `onbuild` functions would work the same as they did before.

---

About the `// TODO: need to check where the SG is being built, NOT where player is? - AfterShock`, it might be too late to address that, as it has the potential to affect how nobuild triggers work on existing maps. However, I'm not sure how much it would actually affect things, so maybe it'd be fine.

If it wouldn't cause problems, it would be possible to still integrate it with the implementation above by moving the construction of the `CFFBuildableInfo` up and using it to get the build origin.

YoYo178:
> I think this can be done in a more backwards compatible way:
> 
> Instead of calling `onbuild` and passing in the buildable entity, `onbuild` could instead be called with an additional a parameter that contains info about the buildable (`m_iWantBuild`, `m_iDetpackTime`, etc; this could either be a plain Lua table with keys for each value (could use `luabind::adl::object` I think) or a new class that's exposed to Lua that contains this info)
> 
> * The current `if( IsInNoBuild() && ... )` conditional would just become `if( IsInNoBuild() )`
> * The values used by `m_iWantBuild` (`FF_BUILD_DISPENSER`, etc) would probably need to be exposed to Lua (see [here](https://github.com/fortressforever/fortressforever/blob/4d8fb70b0ebcaddff59f7d3de0dc3a890ad4909c/dlls/ff/ff_lualib_constants.cpp#L266-L275) for an example)
> * The `IsInNoBuild` function would need to use a custom version of `FFScriptRunPredicates` that can provide more than just the `pEntity` in the function context, maybe a version that takes a `CFFLuaSC*` instead of a `CBaseEntity*` that the current `FFScriptRunPredicates` version could then call to avoid some code duplication.
> * the `CFFLuaSC` that `IsInNoBuild` would want to use would look something like `CFFLuaSC hOutput( 2, (CBaseEntity*)this, eitherLuabindAdlObjectOrSomeLuaBuildableInfoClass );`
> 
> With this implementation, from the Lua side it could look something like this to disallow detpacks with a certain fuse time:
> 
> ```lua
> function your_trigger:onbuild(player, buildable_info)
>   if buildable_info.type == BuildableTypes.kDepack and buildable_info.fusetime < 30 then
>     return false
>   end
>   return true
> end
> ```
> 
> and all existing `onbuild` functions would work the same as they did before.

uhhh, I might try doing something like that but it would be kinda hard for me because I got very less experience about C++

> About the `// TODO: need to check where the SG is being built, NOT where player is? - AfterShock`, it might be too late to address that, as it has the potential to affect how nobuild triggers work on existing maps. However, I'm not sure how much it would actually affect things, so maybe it'd be fine.
> 
> If it wouldn't cause problems, it would be possible to still integrate it with the implementation above by moving the construction of the `CFFBuildableInfo` up and using it to get the build origin.

I did check like 380 Lua files (including the base scripts) which is all I have and didn't find any `nobuild` trigger that's a part of a map's objective (like planting detpacks but not anything else), the only `nobuild` triggers I found was in the spawn rooms of the maps, along with `noannoyances`, so I guess it wouldn't make much of a difference?

squeek502:
> uhhh, I might try doing something like that but it would be kinda hard for me because I got very less experience about C++

No worries. If you don't have any luck with it, I might try implementing what I described to verify that it makes sense.

> I did check like 380 Lua files (including the base scripts) which is all I have and didn't find any `nobuild` trigger that's a part of a map's objective (like planting detpacks but not anything else), the only `nobuild` triggers I found was in the spawn rooms of the maps, along with `noannoyances`, so I guess it wouldn't make much of a difference?

The worry would be that changing `nobuild` to check where the buildable is being built rather than where the player is building from (which is what it does currently) would change where buildables are able/not able to be built, and that some maps might rely on the current behavior. It would definitely be better overall to check where the buildables are trying to be built (most people would probably assume that's how it works), but potentially breaking existing maps is usually worth avoiding. I'd have to check some maps to see if it changes much in practice, though--it's definitely possible that it'd just make things better without any bad results.

YoYo178:
> > uhhh, I might try doing something like that but it would be kinda hard for me because I got very less experience about C++
> 
> No worries. If you don't have any luck with it, I might try implementing what I described to verify that it makes sense.

alright, cool

> > I did check like 380 Lua files (including the base scripts) which is all I have and didn't find any `nobuild` trigger that's a part of a map's objective (like planting detpacks but not anything else), the only `nobuild` triggers I found was in the spawn rooms of the maps, along with `noannoyances`, so I guess it wouldn't make much of a difference?
> 
> The worry would be that changing `nobuild` to check where the buildable is being built rather than where the player is building from (which is what it does currently) would change where buildables are able/not able to be built, and that some maps might rely on the current behavior. It would definitely be better overall to check where the buildables are trying to be built (most people would probably assume that's how it works), but potentially breaking existing maps is usually worth avoiding. I'd have to check some maps to see if it changes much in practice, though--it's definitely possible that it'd just make things better without any bad results.

yeah, I hope it doesn't negatively affect the current maps.

squeek502:
Alright, I ended up implementing what I was thinking of and pushed it to this branch. From the commit message:

> Instead of changing the type of the parameters, the extra context is passed as a new second argument to the `onbuild` callback. The new second argument is a table with the keys `type` that contains a `BuildableTypes` constant and `fusetime` that contains a number with the fusetime of the detpack that is trying to be built (or nil if it's not a detpack being built).
>
> Doing it this way also makes it easy to add more context in the future in a backwards compatible way (just add more stuff to the table).
>
> Example usage:
>
> ```lua
> function nobuild:onbuild( player, buildable_info )
>   -- only allow detpacks if they have a fuse time of exactly 20
>   if buildable_info.type == BuildableTypes.kDetpack and buildable_info.fusetime == 20 then
>     return EVENT_ALLOWED
>   end
>   return EVENT_DISALLOWED
> end
> ```

Also, it turns out the previous code was already simulating the position of the buildable when doing `onbuild` checks, but not using the same code as the actual buildable:

https://github.com/fortressforever/fortressforever/blob/4d8fb70b0ebcaddff59f7d3de0dc3a890ad4909c/dlls/ff/ff_player.cpp#L6871-L6877

I updated that to use `CFFBuildableInfo::GetBuildOrigin` so it now checks the real position where the buildable is trying to be built, which should just be a straightforward improvement (i.e. no real risk of breaking maps, I didn't realize that it wasn't using the position of the player directly previously so I was worrying over nothing before).

PR #396, opened 2024-01-01
+grenade1 and +grenade2 leftovers

BreakinBenny:
Removing some leftovers that aren't ever used. (Must be remnants from TF2's development.)


squeek502:
Thanks!

Some context: FF uses its own `+gren1` and `+gren2` commands defined here:

https://github.com/fortressforever/fortressforever/blob/ede19aa5edef95ffecf8c7964861137fde00e6b8/game_shared/ff/ff_playercommand.cpp#L169-L251

PR #397, opened 2024-01-06
Lua: Add functionality to disable and enable Pyro's jetpack

mvoolt:
Adds CFFPlayer:SetJetpackState(boolean) and CFFPlayer:GetJetpackState()

Made this so I can disable Pyro's jetpack when carrying flag but this can also be useful in other scenarios like in trimp_trainingz_b2 where the map is made before Pyro got a jetpack, Users could just fly instead of actually doing the trimp courses in the map


mvoolt:
Converting to draft because I'm not really sure if the datatable changes are really needed. I'll get back to this tomorrow when I wake up.

mvoolt:
~~Decided to do it anyway, I probably shouldn't have bothered with this but since this is a server-only modification I didn't really see why I should change the DataTables~~
EDIT: I just realized not sending the NetProp value makes no sounds on the Jetpack for clients so whoops! First time programming on the Source SDK.

